PR #1, opened 2017-08-28
Added Meson build definitions.

jpakkane:
For fun and comparison's sake, here are the build definitions in Meson.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,8 @@
+project('c++ starter project', 'cpp',
+  default_options : ['warning_level=3',
+                     'cpp_std=c++14'])
+
+executable('intro', 'main.cpp')
+
+tester = executable('tester', 'tester.cpp')
+test('Tester', tester)


meson.build
```diff
@@ -0,0 +1,8 @@
+project('c++ starter project', 'cpp',
+  default_options : ['warning_level=3',
```
lefticus: How does `warning_level` relate back to `-Wall` `-Wextra` etc, on GCC et al?

jpakkane: Roughly:

    1: -Wall
    2: -Wall -Wextra
    3: -Wall -Wextra -Wpedantic

In addition we also add `-Winvalid-pch`.

We don't have a 4, but there has been discussion on adding one and it would correspond to `-Weverything` on Clang but GCC does not seem to have a preset of this type.

lefticus: Some way to set `/W4` on Windows would be nice via a built-in, since that is what I like to recommend for MSVC.

PR #3, opened 2018-05-08
Added a small getting started with Qt application

paulbendixen:
Well you wanted good getting started with GUI and CMake examples.
This was thrown together in about an hour and a half and shows how easy it is to get started with one of the most powerful GUI libraries out there.
I had to comment out parts of your CMakeLists.txt in order to get my CMake to work, but you could reject those changes.
I've tried to comment as well as I could the thoughts behind the code, I am no Qt expert, but you don't have to be to make something that beats a gtk derivative ;)

This example also relies on system libraries for Qt5, but I don't believe conan has been updated to include Qt yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,18 +17,25 @@ else()
   target_compile_options(project_warnings INTERFACE -Wall -Wextra -Wpedantic)
 endif()
 
-add_executable(intro main.cpp)
+#qt
+find_package( Qt5Widgets REQUIRED )
+set( CMAKE_AUTOMOC ON )
+add_executable( helloQt qt/qtMain.cpp qt/HelloQt.cpp )
+target_link_libraries( helloQt Qt5::Widgets )
+target_compile_features( helloQt PUBLIC cxx_nullptr )
+
+add_executable(intro main.cpp )
 target_compile_features(intro PRIVATE cxx_lambda_init_captures)
 target_link_libraries(intro PRIVATE project_warnings)
 
-target_link_libraries(intro --coverage)
+#target_link_libraries(intro --coverage)
 
 enable_testing()
 
 add_executable(tester tester.cpp)
 target_link_libraries(tester PRIVATE project_warnings)
 
-target_link_libraries(tester --coverage)
+#target_link_libraries(tester --coverage)
 add_test(Tester tester)
 
 
--- a/qt/HelloQt.cpp
+++ b/qt/HelloQt.cpp
@@ -1,0 +1,45 @@
+#include "HelloQt.hpp"
+#include <QHBoxLayout>
+#include <QVBoxLayout>
+#include <QPushButton>
+#include <QLabel>
+#include <QDebug>
+
+HelloQt::HelloQt( QWidget* parent )
+	:QWidget( parent )
+{
+	// The memory management is not done explicitly but by using the parent relationship built into Qt
+	// The tr calls are not strictly neccesary, but it is all that is needed for future i18n.
+	auto mainLayout = new QVBoxLayout;
+	auto descriptiveLabel = new QLabel( tr("Descriptive") );
+	auto pushButton = new QPushButton( tr("Push me!") );
+	auto grouping = new QHBoxLayout;
+	auto directButton = new QPushButton( tr( "Directly!" ) );
+	auto fatal = new QPushButton( tr( "Don't push this button" ) );
+	// By adding the widgets to the main widget, they get parented and the parenting takes care of freeing the memory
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -17,18 +17,25 @@ else()
...
+#target_link_libraries(intro --coverage)
 
 enable_testing()
 
 add_executable(tester tester.cpp)
 target_link_libraries(tester PRIVATE project_warnings)
 
-target_link_libraries(tester --coverage)
+#target_link_libraries(tester --coverage)
 add_test(Tester tester)
```
alexallmont: I recommend putting a CMakeLists.txt into your /qt folder as there can be a lot of options for different platforms. Please feel free to copy the CMakeLists.txt from my minimal project at https://github.com/alexallmont/MinimalQt/blob/master/CMakeLists.txt. This prevent dialogs from popping up in OSX and ensures all the Qt dlls are deployed against the generated exe.

qt/HelloQt.cpp
```diff
@@ -0,0 +1,45 @@
+#include "HelloQt.hpp"
+#include <QHBoxLayout>
+#include <QVBoxLayout>
+#include <QPushButton>
+#include <QLabel>
+#include <QDebug>
+
+HelloQt::HelloQt( QWidget* parent )
+	:QWidget( parent )
+{
```
alexallmont: The style commonly used in Qt is to have tight brackets, e.g. (QWidget* parent) instead of ( QWidget* parent )
They also tend to put the asterisk on the value side in Qt examples, but I agree with you here it should be on the type.

qt/HelloQt.cpp
```diff
@@ -0,0 +1,45 @@
...
+	mainLayout->addWidget( fatal );
+	setLayout( mainLayout );
+	// Signals and slots are use to simply define relationships between actions and effects
+	// By using this "new style" connect the types are checked at compile time to match
+	connect( pushButton, &QPushButton::clicked, this, &HelloQt::writeToDebug );
+	// Lambdas can also be used as can ordinary freestanding functions
+	connect( directButton, &QPushButton::clicked, [](){ qDebug() << tr( "I am in line" ); } );
+	// Most usefull signals and slots are already built into the framework.
+	connect( fatal, &QPushButton::clicked, this, &HelloQt::close );
+};
```
alexallmont: Really nice example of using modern signal convention and use of lambdas and functions. Note that the &HelloQt:: is redundant, you can just use 'close'

qt/HelloQt.hpp
```diff
@@ -0,0 +1,12 @@
+#include <QWidget>
+
```
alexallmont: Some notes on Qt style from their examples:
- Use spaces instead of tabs
- 'public' and 'private slots' should start at column 1, so lines 6-10 should be de-indented by one

qt/HelloQt.hpp
```diff
@@ -0,0 +1,12 @@
...
+class HelloQt : public QWidget
+{
+	Q_OBJECT // Very Important, this is all the magic that is needed
+	public:
+		explicit HelloQt( QWidget* parent = nullptr );
+		~HelloQt();
+	private slots:
+		void writeToDebug( void );
+
```
alexallmont: No need to specify explicit on line 7 or void on line 10

qt/HelloQt.hpp
loomismeister: I really think style guide should be consistent across the cpp_starter_project. If you were going to pick Qt's style guide here, you should consistently use that same style in the other examples.

Better to use the style that is already being used in the rest of the starter project in my opinion.

lefticus: Just run clang-format with your favorite IDE/editor. There's .clang-format file in the top level. There's also a .cmake-format.yml file that can be used to apply appropriate formatting to the CMakeLists.txt

lefticus:
We should change this to pull into the gui branch, to minimize any conflicts in the future.

lefticus:
On second thought, I'll merge the add_gui branch into master.

qt/HelloQt.cpp
paulbendixen: I think I'll let it stand so that the general idea is conveyed

qt/HelloQt.hpp
alexallmont: Sorry Paul, in retrospect I think you're right about explicit usage here. I've just been following the style in the Qt examples, but after a bit of googling the general (non Qt-specific) advice is to use explicit with constructors with one argument.

PR #6, opened 2018-05-14
Add qt gui

paulbendixen:
Updated using the recommendations from @alexallmont.
Remarks:
I haven't run clang-format on the code. The options that are in the .clang-format file are too new for my clang-format-3.5
You should really update your CMake Required version to 3.8, that is when cxx_std_17 makes it into target_compile_features, otherwise use the features you are actually using.
CMakeLists separated into subdirectory as suggested, I also took the liberty of adding some CMake define options so that I could actually run CMake without installing gtk.

### DIFF
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -1,0 +1,6 @@
+find_package( Qt5Widgets REQUIRED )
+set( CMAKE_AUTOMOC ON )
+add_executable( helloQt qtMain.cpp HelloQt.cpp )
+#set_target_properties( helloQt PROPERTIES CMAKE_AUTOMOC ON )
+target_link_libraries( helloQt Qt5::Widgets )
+target_compile_features( helloQt PUBLIC cxx_nullptr cxx_lambdas )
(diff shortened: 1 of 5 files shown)


qt/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+find_package( Qt5Widgets REQUIRED )
+set( CMAKE_AUTOMOC ON )
+add_executable( helloQt qtMain.cpp HelloQt.cpp )
+#set_target_properties( helloQt PROPERTIES CMAKE_AUTOMOC ON )
```
alexallmont: This commented-out line can be removed with CMAKE_AUTOMOC  on

alexallmont:
Great work Paul. It's building fine on my Mac, and I like your use of conditionals so one can focus on a preferred library. It might be handy for build server stages too. @lefticus may need extra README notes to specify that you now have to enable with `cmake -D CPP_STARTER_USE_QT=1 ..` (or _FLTK, _GTKMM etc).

lefticus:
@alexallmont I'll merge it then add option variables for the various GUI components. Thanks everyone!

PR #10, opened 2019-06-17
Use gcc-7 instead of gcc-5 in .travis.yml file

ddalcino:
The last few builds of this project have failed, because the .travis.yml file specified gcc-5 as its compiler, although the CMakeLists.txt file specified several compiler warning flags that were not available in that compiler. Since gcc-7 supports these flags, I elected to use that version instead.


ddalcino:
No longer needed; master branch is now using gcc-10

PR #12, opened 2019-06-17
Add googletest and catch2 testing frameworks

ddalcino:
Hey, I don't know if you're at all interested in these, but I added support for the GoogleTest and Catch testing frameworks. There's a lot of boilerplate CMake code and travis.yml configuration involved in getting these frameworks to work, and I manage to screw it up all the time. This is useful for me personally, so I figure it might be useful for someone else too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ option(CPP_STARTER_USE_FLTK "Enable compilation of FLTK sample" OFF)
 option(CPP_STARTER_USE_GTKMM "Enable compilation of GTKMM sample" OFF)
 option(CPP_STARTER_USE_IMGUI "Enable compilation of ImGui sample" OFF)
 option(CPP_STARTER_USE_NANA "Enable compilation of Nana GUI sample" OFF)
+option(CPP_STARTER_USE_GOOGLE_TEST "Enable Google Test Framework" OFF)
 
 # Link this 'library' to use the following warnings
 add_library(project_warnings INTERFACE)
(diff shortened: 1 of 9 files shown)


mikeloomisgg:
Is it wise to have cmake download the catch repo every time it needs to build? 

Catch2 has a cmakelists install target so that you can install it to cmake's module path even though its header only.
Perhaps it would be better to run the cmake install on catch after first download to install it locally, then cmake would skip the git fetch step.

ddalcino:
You are correct. CMake should not download the catch repo every time; it's a large payload, and the download is very wasteful. However, I am terrible at writing CMakeLists, and this is my best attempt at making it all work on my machine and the CI targets.

To tell the truth, I have never used Catch before. I included it because it felt wrong to support Google Test and not any of the other test frameworks. I will take a look at the Catch2 cmakelists install target, and see if I can get that to work.

Thanks for your input!

mikeloomisgg:
https://github.com/catchorg/Catch2/blob/master/docs/cmake-integration.md#installing-catch2-from-git-repository

There is the documentation for installing catch for use with cmake projects.

I haven't used 'FetchContent' so I'm not totally sure how you'd integrate that advice here. I've always manually installed catch beforehand but it would be nice to have it auto install in a concise way like this.

ddalcino:
Well ... I decided that the best cmake code is no cmake code, so I ripped out the FetchContent stuff and replaced it with a find_package(). I added the 'install catch2' bash script you linked (thanks btw), which works really well on Linux, but it has some issues on Appveyor's Windows environment. Cmake's find_package() requirement appears to be fulfilled, but the compiler can't find <catch2/catch.cpp> for some reason.

Any ideas?

CMakeLists.txt
```diff
@@ -60,56 +61,68 @@ target_link_libraries(intro PRIVATE project_warnings --coverage)
...
+          PUBLIC gtest gtest_main)
+  add_test(NAME GoogleTester COMMAND gtester)
+
+elseif(CPP_STARTER_USE_CATCH2)
+  find_package(Catch2 REQUIRED)
+  add_executable(
+          catch_tester
+          catch_tests/catch_main.cpp
+          catch_tests/catch_test_example.cpp)
+  target_link_libraries(catch_tester
```
mikeloomisgg: Wherever you want the executable to have catch2 include path brought in, you need to link the Catch2::Catch2 library.

mikeloomisgg:
You'll need to include the Catch2::Catch2 library in the target_link_libraries command for any executable that wants to include the catch headers.

CMakeLists.txt
mikeloomisgg: Like in this example https://github.com/catchorg/Catch2/blob/master/docs/cmake-integration.md#usage-1

appveyor.yml
```diff
@@ -1,16 +1,31 @@
...
+    - CMAKE_FLAGS: "-DENABLE_COVERAGE:BOOL=TRUE"
+    - CMAKE_FLAGS: "-DENABLE_COVERAGE:BOOL=TRUE -DCPP_STARTER_USE_GOOGLE_TEST=ON"
+    - CMAKE_FLAGS: "-DENABLE_COVERAGE:BOOL=TRUE -DCPP_STARTER_USE_CATCH2=ON"
+
+before_build:
+  - git clone https://github.com/catchorg/Catch2.git
+  - mv Catch2 _Catch2
+  - cd _Catch2
+  - cmake -Bbuild -H. -DBUILD_TESTING=OFF
+  - cmake --build build/              #--target install # not on Windows!
```
mikeloomisgg: --target install needs to be included, this is where it puts the headers in a user folder and exports the library so that cmake's find_package will find it.

mikeloomisgg: The install will fail without sudo/admin privileges.

CMakeLists.txt
ddalcino: You are correct that I should have added the target_link_libraries() command. I had it in at one point, but I took it out while I was debugging something and forgot to put it back in. 

However, I'm pretty sure that isn't causing the problem with Appveyor. As I understand it, target_link_libraries() affects the link stage. The Visual Studio compiler is won't even produce object files for my catch tests, so the link stage is never reached.

appveyor.yml
ddalcino: I think appveyor does give you admin priveleges. When I used target --install before, the Appveyor VM made a directory in Program Files called Catch2. CMake thinks it can find Catch2 at that point.  However, the compiler could not.

ddalcino: Hey, it passed! Thanks for the help!

ddalcino:
Since Catch2 support has already been added to the project, I no longer think this PR is relevant, so I'm going to close it.

I still think that adding Google Test support would be useful for a lot of people, but it would make a lot more sense to do that as a separate PR.

Issue #16, opened 2020-02-25
Google Test framework support


ddalcino:
I think it would be useful to add support for the Google Test framework, in addition to the project's current support for the Catch2 framework.

PR #17, opened 2020-02-26
Add support for Google Test Framework

ddalcino:
As requested by my issue, #16


ddalcino:
I don't think this is particularly necessary any more. Catch2 is a much more comprehensive and flexible testing framework, and it's probably more useful to encourage people to use it instead.

It's easy enough to install Google Test using Conan yourself if you need it. I will leave this branch as an example to others who need to use Google Test, but I don't think this needs to be merged.

PR #18, opened 2020-02-27
Fix compiler detection for AppleClang and compiler flags for catch_ma…

valerioformato:
Problems I found when I tried to build on OSX:

1. No `-std` flag was passed to the compiler when building `catch_main.cpp`
2. CMake was mistakenly using the sets of warnings defined for gcc. This is because the `CMAKE_CXX_COMPILER_ID` on OSX is actually `AppleClang` and the check for Clang in `CompilerWarnings.cmake` silently fails and fallsback to gcc.

also
3. conan gives an error in the configure phase as seen here: https://github.com/conan-io/cmake-conan/issues/159 . One solution that worked for me was to manually set 
```
export SDKROOT=$(xcodebuild -version -sdk macosx Path)
```
but I don't know if this is "good enough" solution or not.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -79,7 +79,7 @@ function(set_project_warnings project_name)
 
   if(MSVC)
     set(PROJECT_WARNINGS ${MSVC_WARNINGS})
-  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  elseif(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
     set(PROJECT_WARNINGS ${CLANG_WARNINGS})
   else()
     set(PROJECT_WARNINGS ${GCC_WARNINGS})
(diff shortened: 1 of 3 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -79,7 +79,7 @@ function(set_project_warnings project_name)
 
   if(MSVC)
     set(PROJECT_WARNINGS ${MSVC_WARNINGS})
-  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  elseif(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
```
dhruvkakadiya: Same fix is also needed [here](https://github.com/lefticus/cpp_starter_project/blob/d7a2b6b891d75edc68c9d88dca7e24cd73e378b8/cmake/Sanitizers.cmake#L4).

valerioformato: Good catch! Fixed in next commit :)

PR #26, opened 2020-03-03
Disable not compatible sanitizers

davex98:
Together with @robertoleksy


cmake/Sanitizers.cmake
```diff
@@ -35,6 +35,24 @@ function(enable_sanitizers project_name)
...
+
+if (ENABLE_SANITIZER_THREAD AND (ENABLE_SANITIZER_ADDRESS OR ENABLE_SANITIZER_MEMORY))
+    message(FATAL_ERROR "ENABLE_SANITIZER_THREAD is not compatible with "
+        "ENABLE_SANITIZER_ADDRESS or SANITIZER_MEMORY.")
+endif ()
+
+
+if (ENABLE_SANITIZER_MEMORY AND (ENABLE_SANITIZER_THREAD OR ENABLE_SANITIZER_ADDRESS))
+    message(FATAL_ERROR "ENABLE_SANITIZER_MEMORY is not compatible with "
+        "ENABLE_SANITIZER_ADDRESS or SANITIZER_MEMORY.")
```
lefticus, line 54: @davex98 I think this should read "ENABLE_SANITIZER_ADDRESS or SANITIZER_THREAD"?

menuet: Couldn't we do something equivalent but less redundent?
if (ENABLE_SANITIZER_ADDRESS)
    if (ENABLE_SANITIZER_THREAD OR ENABLE_SANITIZER_MEMORY)
        message(FATAL_ERROR "SANITIZER_ADDRESS is not compatible with "
        "SANITIZER_THREAD or SANITIZER_MEMORY.")
    endif ()
elseif (ENABLE_SANITIZER_THREAD AND ENABLE_SANITIZER_ADDRESS)
    message(FATAL_ERROR "ENABLE_SANITIZER_THREAD is not compatible with SANITIZER_MEMORY.")
endif ()

ddalcino:
These changes appear to be superseded by #68, which was merged 5 months ago.  

There appears to be some disagreement about what should happen when multiple sanitizers are enabled. In #68, multiple sanitizers cause a `WARNING` message; here it's a `FATAL_ERROR`. I think that there's some value in changing the error message to a `FATAL ERROR`, but I think that would make more sense as a separate PR.

Should this PR be closed?

lefticus:
Thanks @ddalcino I'm closing this now.

PR #30, opened 2020-03-04
Fix mis-detection of Clang on macOS

acgetchell:
It is better to test for "GNU" rather than "Clang" because on macOS that will resolve to "AppleClang" and the GNU warnings will be enabled and produce errors.


valerioformato:
I believe this has already been addressed by #18 

Also... with that final `else` statement the Clang setting would be applied also if one tries to use ICC or some other compiler, I don't know if they would work in that case 🤔

lefticus:
Already addressed in #18, closing. Sorry for the late review!

PR #35, opened 2020-03-11
Add a working Mac CI build

ddalcino:
There are a lot of people hanging out in the issues section that are having trouble getting this project to build properly on a Mac. For this reason, I think that it's really important to get continuous integration going on the Mac side. This will prevent a lot of headaches for Mac users.

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -18,7 +18,7 @@ conan_cmake_run(
   ${CONAN_EXTRA_REQUIRES}
   catch2/2.11.0
   docopt.cpp/0.6.2
-  fmt/6.0.0
+  fmt/6.1.2
   spdlog/1.5.0
   OPTIONS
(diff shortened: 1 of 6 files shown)


cmake/Conan.cmake
```diff
@@ -18,7 +18,7 @@ conan_cmake_run(
   ${CONAN_EXTRA_REQUIRES}
   catch2/2.11.0
   docopt.cpp/0.6.2
-  fmt/6.0.0
+  fmt/6.1.2
```
ddalcino: This looks like a strange thing to change, but the Mac build consistently fails when fmt is set to 6.0.0. It also fails on my Fedora box. Updating fmt to 6.1.2 appears to correct the issue.

See this issue for reference: https://github.com/lefticus/cpp_weekly_game_project/issues/3

lefticus:
@ddalcino thank you!

PR #36, opened 2020-03-11
Add basic documentation

ddalcino:
This repo would be a lot more beginner-friendly if it had some helpful instructions on how to build it. It seems to me like there are a lot of people having trouble building this project; having some documentation like this would go a long way towards alleviating their issues.

Regarding what I've written: Maybe it's too beginner-focused (the first section explains how to use git). Personally, I don't think it's beginner-focused enough. We want people to have an easier time getting started with C++, and having good documentation will help do that.

Even if you're not a beginner, it would be nice if the README told you up front that you need to have Conan installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,14 @@ frameworks (fltk, gtkmm, imgui, etc.), you can remove them with `git rm`:
 ## Dependencies
 
 ### Necessary Dependencies
-1. A C++ compiler that supports C++17 (gcc 7+, clang 6+, Visual Studio 2019)
+1. A C++ compiler that supports C++17. 
+See [cppreference.com](https://en.cppreference.com/w/cpp/compiler_support)
+to see which features are supported by each compiler.
+The following compilers should work:
+  * [gcc 7+](https://gcc.gnu.org/)
+  * [clang 6+](https://clang.llvm.org/)
+  * [Visual Studio 2017](https://visualstudio.microsoft.com/) or higher. 
+  The minimum compiler version is 19.15; this version ships with version 15.8 of the IDE. 
 2. [Conan](https://conan.io/) - it's recommended that you install using 
 [pip](https://pip.pypa.io/en/stable/) 
 3. [CMake](https://cmake.org/)
--- a/README.md
+++ b/README.md
@@ -152,3 +152,10 @@ earlier version of `fmt`, such as `fmt/6.0.0`, and then run:
 
 ...the first line of output would be a warning that `spdlog` needs a more recent
 version of `fmt`.
+
+## Testing
+See [Catch2 tutorial](https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md)
+
+## Fuzz testing
+
+See [libFuzzer Tutorial](https://github.com/google/fuzzing/blob/master/tutorial/libFuzzerTutorial.md)
--- a/README.md
+++ b/README.md
@@ -25,24 +25,37 @@ frameworks (fltk, gtkmm, imgui, etc.), you can remove them with `git rm`:
 
     $ git rm -r src/<unnecessary_framework>
 
-## Necessary Dependencies
+## Dependencies
 
+### Necessary Dependencies
 1. A C++ compiler that supports C++17 (gcc 7+, clang 6+, Visual Studio 2019)
 2. [Conan](https://conan.io/) - it's recommended that you install using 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,3 +5,62 @@
 [![Build Status](https://travis-ci.org/lefticus/cpp_starter_project.svg?branch=master)](https://travis-ci.org/lefticus/cpp_starter_project)
 
 [![Build status](https://ci.appveyor.com/api/projects/status/ro4lbfoa7n0sy74c/branch/master?svg=true)](https://ci.appveyor.com/project/lefticus/cpp-starter-project/branch/master)
+
+
+## Getting Started
+
+
+    $ mkdir <project_directory_name>
```
ddalcino: This stuff is repeat from the C++ Weekly video, but it's helpful to have it written out all in one place.
It would be completely unnecessary if this were a template repository.

README.md
```diff
@@ -5,3 +5,62 @@
...
+    $ git remote add origin https://github.com/<user>/<your_new_repo>.git
+
+If you know you're not going to use one or more of the optional gui/graphics 
+frameworks (fltk, gtkmm, imgui, etc.), you can remove them with `git rm`:
+
+    $ git rm -r src/<unnecessary_framework>
+
+## Necessary Dependencies
+
+1. A C++ compiler that supports C++17 (gcc 7+, clang 6+, Visual Studio 2019)
```
ddalcino: I'm not certain which compilers are actually required for this project; this info is coming from https://en.cppreference.com/w/cpp/compiler_support. In any case, there should probably be separate CI builds for all the compilers listed here, and currently there are not.

README.md
```diff
@@ -5,3 +5,62 @@
...
+    are using. Please see the documentation for your particular framework for 
+    details:
+    
+    - [FLTK](https://www.fltk.org/doc-1.4/index.html)
+    - [GTKMM](https://www.gtkmm.org/en/documentation.html)
+    - [IMGUI](https://github.com/ocornut/imgui/tree/master/docs)
+    - [NANA](http://nanapro.org/en-us/documentation/)
+    - [QT](https://doc.qt.io/)
+    - [SDL](http://wiki.libsdl.org/FrontPage)
+    - [SFML](https://www.sfml-dev.org/tutorials/2.5/compile-with-cmake.php)
```
ddalcino: I'm not sure that this section should be here. It makes it look like you have to install all of these frameworks independently.

This section is included because I had some trouble getting IMGUI running on my Fedora box: I had to install libudev and xrandr using my package manager. The SFML documentation helped me figure that out.

In case anybody else has the same trouble, the libudev would not link properly on Fedora 30 until I installed the 'systemd-devel' package, strangely enough.

lefticus: This looks mostly good to me. The only issue (I think) is that SDL and SFML should be pulled in with Conan and I don't think those deps need to be listed?

I'm going to merge as-is though.

README.md
lefticus: We do need to enhance the CI, but this looks good right now.

README.md
```diff
@@ -5,3 +5,157 @@
...
+* Tell CMake which compiler to use, via the CMake variables CMAKE_C_COMPILER 
+  and CMAKE_CXX_COMPILER
+
+Be aware that CMake will detect which compiler was used to build each of the 
+Conan targets, and if it doesn't match CMAKE_C_COMPILER or CMAKE_CXX_COMPILER,
+the project will fail to build.
+
+To build using clang, you can use these commands:
+
+    $ CC=clang CXX=clang++ ccmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ ..
```
lefticus, line 98: I thought `CMAKE_C_COMPILER` would default to `CC`

lefticus: Yes: https://cmake.org/cmake/help/v3.10/envvar/CC.html

ddalcino: You're right. I think I meant to change this at one point but never got around to it. Thanks for catching this.

README.md
ddalcino: I recently learned that I was misreading the Visual Studio version numbers, and that Visual Studio 2017 (version 15.8 with compiler 19.15) actually fully supports C++17, so I'm going to change the version number on that dependency accordingly.

lefticus:
Thank you for the huge amount of work and back-and-forth on this one!

PR #37, opened 2020-03-14
.travis.yml compilers in CXX and CC were swapped

carlsonmark:
CXX should be g++, and CC should be gcc.
This was the case a month ago, but when they were updated to use gcc 10, they were swapped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
 
 
 script:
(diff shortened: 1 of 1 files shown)


lefticus:
Thank you!

This was fixed by a previous PR that I just merged however, which had much more extensive reworking to the travis config.

PR #39, opened 2020-03-23
Update CompilerWarnings to include permissive-

DeveloperPaul123:
Added flag for standards conforming code in CompilerWarnings on Windows MSVC builds.


lefticus:
Yes!

PR #41, opened 2020-03-28
adding support for apple clang

amitsingh19975:
It did not run mac but it gave errors that Clang does not support -Wmisleading-indentation, -Wduplicated-cond, etc  because cmake sets CMAKE_CXX_COMPILER_ID as "AppleClang" rather than "Clang" so I added the regex to it.


ddalcino:
This appears to be addressed by PR #18

PR #49, opened 2020-04-19
Ignore cppcheck unused function warning for fuzzing entry point

avitase:
cppcheck warning 'unusedFunction' for `LLVMFuzzerTestOneInput` should be ignored. In order to make this work the option `--inline-suppr` has to be added to the cppcheck invocation. While doing this I found that the ENABLE_FUZZING option was missing in ccmake and that the cmake files were not properly formatted according to `.cmake-format.yaml`. Further, cmake-format's option 'max-subargs_per_line' was removed in version 0.6.0.

### DIFF
--- a/.cmake-format.yaml
+++ b/.cmake-format.yaml
@@ -12,7 +12,7 @@ dangle_parens: false
 enum_char: .
 line_ending: unix
 line_width: 120
-max_subargs_per_line: 3
+max_pargs_hwrap: 3
 separate_ctrl_name_with_space: false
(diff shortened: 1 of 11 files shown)


lefticus:
Sorry, the large reformatting of the files (probably necessary) makes it hard to tell which files actually changed.

Can you please reduce your PR to just the change you were making?

avitase:
I split this request into three separate requests and do not run cmake-format. However, at some point one should consider actually applying cmake-format to all cmake files once and for all ...

PR #57, opened 2020-04-26
Update StandardProjectSettings.cmake

giellamoswhard:
I've added colored diagnostic for Clang and Gcc: for me it is really useful.

### DIFF
--- a/cmake/StandardProjectSettings.cmake
+++ b/cmake/StandardProjectSettings.cmake
@@ -37,3 +37,11 @@ if(ENABLE_IPO)
 endif()
 
 
+if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+  add_compile_options(-fcolor-diagnostics)
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


cmake/StandardProjectSettings.cmake
```diff
@@ -37,3 +37,11 @@ if(ENABLE_IPO)
 endif()
 
 
+if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+  add_compile_options(-fcolor-diagnostics)
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  add_compile_options(-fdiagnostics-color=always)
+else()
+  message(AUTHOR_WARNING "No colored compiler diagnostic set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
```
aminya, line 45: This should not be a warning.
```suggestion
else()
  message(STATUS "No colored compiler diagnostic set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
```

giellamoswhard:
Yes. It's Better! Thanks

PR #59, opened 2020-05-01
Add .gitignore file

davidcampelo:
Avoid "git status" to show annoying dirty working copy with generated files. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,50 @@
+# Build Files
+**/CMakeFiles/
+**/CMakeCache.txt
+**cmake_install.cmake
+compile_commands.json
+Makefile
+**CTestTestfile.cmake
+test/**.cmake
+
+# Conan files
+/conan.cmake
+conan.lock
+conanbuildinfo.cmake
+conanbuildinfo.txt
+conanfile.txt
+conaninfo.txt
+
+graph_info.json
+
+# Library files
+lib/
+
+# Binary files
+imgui.ini
+bin/
+build/
+out/
+cmake-build-debug/
+cmake-build-release/
+
+# Annoying OS Generated Files
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+ehthumbs.db
+Thumbs.db
+.directory
+*~
+
+# IDE files
+.vs/
+.idea/
+.vscode/
+language.settings.xml
+**/**.swp
+
+# python venv files
+virtualenv/


avitase:
you should also add vim's *.swp files and the venv/ directory, for those using a virtualenv to install conan and cmake_format.

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+**/CMakeFiles/
+**/CMakeCache.txt
+**cmake_install.cmake
+compile_commands.json
+Makefile
+**CTestTestfile.cmake
+
+# Visual Studio Code project files
+.vscode/
+language.settings.xml
```
aminya, line 11:
```suggestion
# IDE files
.vs/
.idea/
.vscode/
language.settings.xml
```

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+
+# Libary files
+lib/
+
+# Binary files
+imgui.ini
+bin/
+build/
+test/
+
```
aminya, line 31:
```suggestion
out/
```

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+language.settings.xml
+
+# Conan files
+conan.cmake
+conan.lock
+conanbuildinfo.cmake
+conanbuildinfo.txt
+conanfile.txt
+conaninfo.txt
+
```
aminya, line 20:
```suggestion
virtualenv/
```

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+graph_info.json
+
+# Libary files
+lib/
+
+# Binary files
+imgui.ini
+bin/
+build/
+test/
```
aminya, line 30: You should not gitignore the test folder!
```suggestion

```

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+compile_commands.json
+Makefile
+**CTestTestfile.cmake
+
+# Visual Studio Code project files
+.vscode/
+language.settings.xml
+
+# Conan files
+conan.cmake
```
aminya, line 14: You should not gitignore Conan.cmake. All the deps live here!!
```suggestion

```

davidcampelo: conan.cmake file is downloaded from the repo automatically the first time cmake runs. it's not part of the project.

aminya: There is another file called Conan.cmake under cmake folder. This conflicts with that.

davidcampelo: thanks for pointing that out. I believe using '/conan.cmake' would fix this.

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+.DS_Store
+.DS_Store?
+._*
+.Spotlight-V100
+.Trashes
+ehthumbs.db
+Thumbs.db
+.directory
+*~
+
```
aminya, line 39:
```suggestion
# IDE files
.vs/
.idea/
.vscode/
```

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+# Libary files
+lib/
+
+# Binary files
+imgui.ini
+bin/
+build/
+out/
+
```
aminya, line 28: Clion build folder
```suggestion
cmake-build-debug/
cmake-build-release/
```

aminya:
@davidcampelo Could you resolve the suggestions that you have applied already?

ddalcino:
Question regarding the first 16 lines: Are they intended for an in-source build? As far as I can tell, the only way for a lot of those files to get into your project root directory is if you built your project there. 

Since this project no longer supports in-source builds, should these lines still be in the .gitignore file?

lefticus:
It's actually because of this PR, which I looked at then failed to comment on, that I disabled in-source builds

Jason5480:
Shall we also ignore binary files in the source repo? The proposed method to use third party tools and executables is through conan packages. So I would add the following lines in .gitignore file
````
# Do not ever commit binaries in a source repo. Create a conan package instead!

# Compiled Object files
*.slo
*.lo
*.o
*.obj

# Precompiled Headers
*.gch
*.pch

# Compiled Dynamic libraries
*.so
*.dylib
*.dll

# Compiled Static libraries
*.lai
*.la
*.a
*.lib

# Executables
*.exe
*.out
*.app
````

File: .gitignore [hunk 8]
```diff
@@ -0,0 +1,50 @@
...
+Makefile
+**CTestTestfile.cmake
+test/**.cmake
+
+# Conan files
+/conan.cmake
+conan.lock
+conanbuildinfo.cmake
+conanbuildinfo.txt
+conanfile.txt
```
aminya, line 15:
```suggestion
```

This conflicts with #65

lefticus:
Because I have disabled in-source builds, I'm going to close this PR.

If you disagree, please reopen and comment.

aminya:
@lefticus I think despite this, having gitignore does not have any disadvantageous. Someone might prefer to use in-source builds.

ddalcino:
@lefticus I am a little disappointed that this didn't get merged in some form either. It's useful to have a `.gitignore` file here, even if it's empty. Its presence will remind users that they need to add things to it.

Personally, I would remove the first 18 lines, to make it doubly clear that in-source builds are not a good idea, and keep the rest. The rest of the file won't be useful for everyone, but I think parts of it are useful to most people.  None of the lines after 18 are likely to conflict with something you would want to commit in a C++ project, and if they do, it's not hard for a user to discover and remove them.

Also, I wanted to thank you for taking the time to maintain this repo. I recognize that it's an often thankless job, and most your interactions with users are with people like me who want you to read/evaluate/merge their PRs, or people in the 'Issues' section who are having trouble. I think we all owe you our thanks and a tall glass of beer.

PR #60, opened 2020-05-02
Add VS Code workspace

### LINKED ISSUE #161: VS Code Quickstart
Two of the oldest open PRs that we have, #60 and #99, are attempting to add support for VS Code. I've never used VS Code for C++ before, so I am not sufficiently qualified to review them. My gut instinct is that IDE-specific code does not belong in this project: this project is about C++, not an IDE. Besides, if something goes wrong with the VS Code code (😃), someone has to debug it.

However, I do understand that it's not so trivial to set up VS Code with an existing C++ project. Since this project works as a batteries-included quick-start for C++ in so many other ways, maybe it would be really helpful to add some batteries for VS Code.

If we do add support for VS Code, I would like to make sure that it also works for VSCodium/Code_OSS/ whatever other telemetry-free alternatives people are using these days.

shreyasbharath:
I got the project building on VS Code (Windows) with the official CMake extension.

Fix #161

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,8 @@
+build/*
+
+.vscode/*
+!.vscode/cpp_starter_project.code-workspace
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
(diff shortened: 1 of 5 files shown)


ddalcino:
Superseded by #175

PR #63, opened 2020-05-20
Fix spdlog dependency warning

ddalcino:
spdlog 1.5 and 1.6 now require fmt 6.2.0, as of this commit:
https://github.com/conan-io/conan-center-index/pull/1651/commits/c19230f8ad6ae31a3699c4a66d90e29bad502ecb

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -18,7 +18,7 @@ conan_cmake_run(
   ${CONAN_EXTRA_REQUIRES}
   catch2/2.11.0
   docopt.cpp/0.6.2
-  fmt/6.1.2
+  fmt/6.2.0
   spdlog/1.5.0
   OPTIONS
   ${CONAN_EXTRA_OPTIONS}


cmake/Conan.cmake
```diff
@@ -18,7 +18,7 @@ conan_cmake_run(
   ${CONAN_EXTRA_REQUIRES}
   catch2/2.11.0
```
aminya, line 19:
```suggestion
  catch2/2.12.1
```

cmake/Conan.cmake
```diff
@@ -18,7 +18,7 @@ conan_cmake_run(
...
   docopt.cpp/0.6.2
-  fmt/6.1.2
+  fmt/6.2.0
   spdlog/1.5.0
```
aminya, line 22:
```suggestion
  spdlog/1.6.0
```

cmake/Conan.cmake
ddalcino: Thanks for the approval and suggestions, but for this PR I was really only interested in fixing the dependency warning. If the developers of spdlog think that their package requires a more up-to-date version of fmt, then there's probably a bug somewhere that's caused by pairing spdlog with the older version. That's the bug that this PR is meant to prevent.

You are welcome to make your own PR to update the other dependencies; that's probably a good idea. I would rather keep those changes separate from this PR though.

PR #65, opened 2020-05-22
conanfile.txt

aminya:
This moves the hardcoded dependencies from Conan.cmake to the conanfile.txt. 

This has many benefits:
- this is the standard way of keeping track of the deps.
- allows using the tools for automatic bumping, etc (they work on conanfile.txt).

Also supersedes https://github.com/lefticus/cpp_starter_project/pull/63


lefticus:
Just to verify - the `Requires` block and the `conan_file` block can coexist? I tend to put the dependencies inside of the CMakeLists so it's easy to make them optional depending on the build options chose.

aminya:
@lefticus It seems to work, but apparently this is not supported officially. 
I have created a couple of issues to track this.
https://github.com/conan-io/cmake-conan/issues/235
https://github.com/conan-io/cmake-conan/issues/236
https://github.com/conan-io/conan/issues/7068

ddalcino:
There's another approach to solving this problem: You can have 1 `conanfile.txt` at the root level, and additional `conanfile.txt`s within each example subdirectory. When that subdirectory's `CMakeLists.txt` runs, it calls `conan_cmake_install` on the local `conanfile.txt`. This moves the dependencies for the examples out of the root level CMake code and into the locations where they are needed.

I have a [branch](https://github.com/lefticus/cpp_starter_project/compare/master...ddalcino:add-conanfiles) that implements this approach, and the examples build successfully for me.

Potential downsides:

- There are now 3 `conanfile.txt`s instead of one, and I'm not sure if Conan explicitly supports this.
- The `conan_cmake_run()` function needs to be called 3 times; I don't think that's inappropriate. It would work better to separate those calls into `conan_cmake_configure()`, `conan_cmake_autodetect()`, and 3 calls to `conan_make_install()`, per the docs at https://github.com/conan-io/cmake-conan#cmake-conan. Apparently, `conan_cmake_run()` is going to be deprecated; see https://github.com/conan-io/cmake-conan#conan_cmake_run-high-level-wrapper.

aminya:
I've been using what I have in this PR for a while, and I haven't had any issues. I prefer one single file to manage the dependencies.

ddalcino:
I just pulled this branch and tried to build the IMGUI example, using Conan version 1.34.0. CMake gives me this error message during configuration:

```
CMake Error at src/imgui/CMakeLists.txt:2 (add_executable):
  Target "test_imgui" links to target "CONAN_PKG::imgui-sfml" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

When I add `imgui-sfml/2.1@bincrafters/stable` to `conanfile.txt`, it builds just fine. Perhaps the Conan devs explicitly removed the ability to mix `REQUIRES` with `CONANFILE`?

I know this used to work, but have you tried it recently?

aminya:
I haven't updated my Conan recently, but the CI passes. Maybe, we should build one of the optional libs in the CI to make sure it works.

ddalcino:
The CI doesn't test any of the GUI examples, but I agree that it should.

aminya:
Closing in favor of #126

PR #67, opened 2020-05-23
Add more ci builds

ddalcino:
This PR adds several extra CI builds for Mac and Linux.

I have added several basic builds to test the following: 
* gcc 7-10 and clang 6&8-10 on Ubuntu 18 (for basic compiler compatibility)
* gcc-9 and apple-clang-11 on Mac OS 10.13, 10.14, and 10.15 (for OSX compatibility)
* A very small subset of the CMake options this project offers, on clang-9 and gcc-9 on Ubuntu 18

I did not go very far with regard to testing different combinations of the CMake options. There are 2^n possible ways to choose these options; complete coverage would be absurd.

Similarly, there's a combinatorial explosion of different possible OS/compiler setups to use, and I'm not going to try for complete coverage there either.

I have noticed that the more CI builds you attempt at once, the higher the chance that a few of them will fail due to package manager timeouts. At this stage, I have to re-trigger several builds manually before they all pass.

I feel that several (if not all) of these CI builds are necessary, just so we know what environments this project does and does not work in. However, I'm not convinced that these extra builds are useful to our users. I would expect most users to rip out most of these builds as soon as they begin their projects.

I am posting this as a draft PR because I need feedback: This is definitely not ready to merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ jobs:
                        -D ENABLE_SANITIZER_MEMORY:BOOL=ON \
                        -D ENABLE_SANITIZER_THREAD:BOOL=ON \
                        -D ENABLE_SANITIZER_UNDEFINED_BEHAVIOR:BOOL=ON"
-      install:
+      before_install:
         - brew install cppcheck
 
     - os: linux     # GCC-7 - Codecov build
@@ -161,7 +161,7 @@ jobs:
                        -D ENABLE_SANITIZER_MEMORY:BOOL=ON \
                        -D ENABLE_SANITIZER_THREAD:BOOL=ON \
                        -D ENABLE_SANITIZER_UNDEFINED_BEHAVIOR:BOOL=ON"
-      install:
+      before_install:
         - eval "${MATRIX_EVAL}"
         - git clone https://github.com/include-what-you-use/include-what-you-use.git
         - cd include-what-you-use && git checkout "clang_${CLANG_SEMVER}"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,57 +1,221 @@
...
+    - os: osx       # 10.15.4 - clang sanitize-all-the-things build
+      compiler: clang
+      osx_image: xcode11.4
+      env:
+        - MACOSX_DEPLOYMENT_TARGET=10.15
+        - CONFIG_OPTS="-D ENABLE_CPPCHECK:BOOL=ON \
+                       -D ENABLE_SANITIZER_ADDRESS:BOOL=ON \
+                       -D ENABLE_SANITIZER_MEMORY:BOOL=ON \
+                       -D ENABLE_SANITIZER_THREAD:BOOL=ON \
+                       -D ENABLE_SANITIZER_UNDEFINED_BEHAVIOR:BOOL=ON"
```
ddalcino, line 56: This kind of test should probably be an EXPECT_FAIL kind of thing. I think we don't really want to run more than one thread sanitizer at once.
We should run all the sanitizers, but in sequence, not simultaneously.

lefticus: Yeah, definitely have to do something about the sanitizers, they are not compatible together. The build will either warn or fail here, because the compiler doesn't allow it (but CMakeLists is supposed to check for it)

ddalcino: This particular build did not fail: https://travis-ci.org/github/lefticus/cpp_starter_project/builds/690480254

I think #26 is meant to address this issue?

ddalcino:
I’m thinking about closing this, even though it might be useful for some people. Travis CI recently added limits to the amount of time that can be spent running CI tests for a free open-source account; this PR will eat up all those minutes real quick. Normally, one build of the project takes around 10 min; this PR takes nearly an hour.

PR #70, opened 2020-05-30
C++17 filesystem on travis, outdated libstdc++ 

extramask93:
Using the `std::filesystem` features caused travis to fail builds on ubuntu and mac with clang giving:

-   `fatal error: 'filesystem' file not found` due to an outdated standard library implementation, which is fixable by adding the newer `libstdc++-10-dev` package in `.travis.yml` 

- `error: 'current_path' is unavailable: introduced in macOS 10.15` which is fixable by introducing `mmacosx-version-min=10.15` into the `CMakeLists.txt`

respectively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,16 +43,17 @@ jobs:
       dist: bionic
       compiler: clang
       env:
-        - CLANG_VER="9"
+        - CLANG_VER="10"
         - MATRIX_EVAL="CC=clang-${CLANG_VER} && CXX=clang++-${CLANG_VER}"
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - llvm-toolchain-9
+            - sourceline: 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main'
+              key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
           packages:
-            - clang-9
-            - libstdc++-9-dev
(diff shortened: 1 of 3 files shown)


lefticus:
I'm agreed that we want std::filesystem support, and full C++17, but this also drastically limits the MacOS compilers allowed. 

In general I'm inclined to accept the PR.

ddalcino:
Since we're moving away from Travis (see #157), do we still need this PR?

It looks like the only change to root level `CMakeLists.txt` just sets a minimum version of MacOS. I am opposed to this. I think that a user on MacOS 10.14 or below, who wants to use C++17 features, would either upgrade their OS or find some alternate means of installing a modern clang toolchain that includes those features.

I am not opposed to adding `std::filesystem` calls to `src/main.cpp`; I think that could be a useful diagnostic to make sure your toolchain is set up properly.

This PR is more than a year old; I think it's time to close it or merge it.

ddalcino:
Ok, it has been more than two weeks since I requested reviews for this PR, and no one appears to be interested. Additionally, the author has not addressed the issues with Travis or the minimum MacOS requirement. Since no one has said anything, I understand this to mean that we do not need this PR, so I'm closing it now.

@extramask93, you are welcome to reopen this PR if you really feel strongly about it, and are willing to remove or alter the changes to `.travis.yml` and `CMakeLists.txt`.

PR #86, opened 2020-08-29
Use python3 virtual environment instead of python2

ddalcino:
### Why:
When Conan version 2.0 is released, it will require Python 3 to run. When that happens, this project's CI builds may break when Conan runs using Python 2.

### How:
This uses `venv` to create a virtual environment that will always use Python version 3 whenever `python` or `pip` are run. This solution works the same way on Linux, Mac, and Windows, and it works no matter what version of Python is the default.

### Why remove the `--user` flag:
After venv has been run, all future invocations of `pip` will apply only to the local virtual environment, and not the system as a whole, so the `--user` flag is not allowed.

### Other possible solutions:
* An alternate and much simpler solution would be to use more recent images of Ubuntu and MacOS that use Python 3 as the default version of Python, for instance, Ubuntu 20.04. I am hesitant to recommend this solution because it breaks compatibility with older versions of Ubuntu and Mac.
* Another possibility would be `update-alternatives`, which is a great way to manage different versions of programs on Ubuntu. To my knowledge, it doesn't exist on Mac.
* Travis appears to use `pyenv` to set the desired version of Python for Python projects. I have been unable to get `pyenv` to install the same version of Python on both Linux and Mac without writing a lot of platform-dependent code.
* Use `virtualenv` instead of `venv`. `virtualenv` is preinstalled on both Linux and Mac travis-ci images, so this would eliminate the need to `apt-get install python3-venv` on Linux. This would work, but `virtualenv` is meant for Python 2 package management. The Python docs specifically recommend using `venv` for package management when using Python 3.
* Use `python3 -m pip install <stuff>` instead of using a virtual environment. This approach works just fine, and the code is simpler. You can see what a working example of this would look like on this branch: https://github.com/ddalcino/cpp_starter_project/tree/use_python3. To be honest, I think this may be a better approach than using a venv. I will update this comment if I can think of any real disadvantage to this approach in a CI environment.

Edit on Sept 7 to add `python -m pip install` alternate approach

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ install:
   - python --version | grep -iE 'python\s*3'
   - pip    --version | grep -iE 'python\s*3'
 
+  # venv manages to find pip v9.0, but version 20.2 is current
+  - pip install --upgrade pip
+
   # Install Conan using Python 3.
   - pip install conan cmake
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,19 @@
 language: cpp
 
 install:
-  - pip install --user conan cmake
+  # Create virtual environment for Python 3
+  - python3 -m venv ~/virtual_env
+  - source ~/virtual_env/bin/activate
+
+  # Fail if we are using the wrong Python
+  - python --version | grep -iE 'python\s*3'
+  - pip    --version | grep -iE 'python\s*3'
+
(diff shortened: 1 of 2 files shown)


ddalcino:
I have decided that the approach in #88 makes more sense than this. Closing #86.

PR #89, opened 2020-09-19
Fix adding compilers to Windows PATH during installation

aminya:
This extra flag commands vswhere to include any product, including VC build tools, in the search. People might not install a full Visual Studio, and this adds support for that.


aminya:
@lefticus Could you take a look at the pull requests I have opened?

aminya:
@lefticus Could you check my PRs?

PR #93, opened 2020-09-30
Fixed -ftime-trace option registration command

hadrabap:
Hello,

I've been playing around a bit with your `cpp_starter_project` and I think I found an issue with how the `-ftime-trace` option is being registered into the project. It seems I've also find a solution, but could you please take a look at it as I'm fairly new to CMake. Thanks in advance.

Cheers,

PETR


lefticus:
@hadrabap looks correct.

PR #94, opened 2020-10-02
Add Linker configuration file

patrickelectric:
Allow user to configure and choose different linkers to improve compilation time, like: lld, gold, bfd and etc.
Reference: https://www.youtube.com/watch?v=yTtWohFzS6s

### DIFF
--- a/cmake/Linker.cmake
+++ b/cmake/Linker.cmake
@@ -1,0 +1,33 @@
+option(ENABLE_USER_LINKER "Enable a specific linker if available" OFF)
+
+include(CheckCXXCompilerFlag)
+
+set(USER_LINKER_OPTION
+  "lld"
+  CACHE STRING "Linker to be used")
+set(USER_LINKER_OPTION_VALUES "lld" "gold" "bfd")
+set_property(CACHE USER_LINKER_OPTION PROPERTY STRINGS ${USER_LINKER_OPTION_VALUES})
+list(
+  FIND
+  USER_LINKER_OPTION_VALUES
+  ${USER_LINKER_OPTION}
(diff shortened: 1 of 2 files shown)


cmake/Linker.cmake
```diff
@@ -0,0 +1,33 @@
...
+function(configure_linker)
+    if(NOT ENABLE_USER_LINKER)
+        return()
+    endif()
+
+    set(LINKER_FLAG "-fuse-ld=${USER_LINKER_OPTION}")
+
+    check_cxx_compiler_flag(${LINKER_FLAG} CXX_SUPPORTS_USER_LINKER)
+    if(CXX_SUPPORTS_USER_LINKER)
+        add_compile_options(${LINKER_FLAG})
```
ddalcino, line 31: Quick question about this: 

Would there be any reason to use `target_compile_options` here instead, so that different targets could use a different linker? 

The `add_compile_options` function is usually a red flag for me, because it tends to make CMake code less flexible. On the other hand, I can't think of any reason you would want to use one linker for one target and a different one for another.

patrickelectric: my fault, moving to `target_compile_options`

patrickelectric:
ping @ddalcino

ddalcino:
> ping @ddalcino

This looks pretty good to me; I definitely appreciate that you used `target_compile_options`. I really like being able to pick a different linker, and this PR makes it easy to do so. If this were my repository, I would merge this. I would also merge a lot of the other open PRs here. However, I am not a CMake/C++ expert, and I do not have any pull with the the repo's owner.

I think you mean to ping @lefticus.

PR #81, opened 2020-08-05
Fix build error on code page 949

raeperd:
When I built without making any changes to this project, I found the following error.

```powershell
[main] Building folder: cpp_starter_project 
[build] Starting build
[proc] Executing command: "C:\Program Files\CMake\bin\cmake.EXE" --build c:/lab/cpp_starter_project/build --config Debug --target ALL_BUILD -- /maxcpucount:14
[build] Microsoft (R) Build Engine version 16.5.0+d4cbfca49 for .NET Framework
[build] Copyright (C) Microsoft Corporation. All rights reserved.
[build] 
[build]   Checking Build System
[build]   Building Custom Rule C:/lab/cpp_starter_project/test/CMakeLists.txt
[build]   Building Custom Rule C:/lab/cpp_starter_project/src/CMakeLists.txt
[build]   catch_main.cpp
[build]   main.cpp
[build] C:\Users\PC\.conan\data\fmt\6.2.0\_\_\package\9ffec4a59d60a2ccf636e06ba57253e21417f01b\include\fmt/core.h(327,35): error C2220: the following warning is treated as an error [C:\lab\cpp_starter_project\build\src\intro.vcxproj]
[build] C:\Users\PC\.conan\data\fmt\6.2.0\_\_\package\9ffec4a59d60a2ccf636e06ba57253e21417f01b\include\fmt/core.h(327,35): warning C4566: character represented by universal-character-name '\u00B5' cannot be represented in the current code page (949) [C:\lab\cpp_starter_project\build\src\intro.vcxproj]
[build] C:\Users\PC\.conan\data\spdlog\1.5.0\_\_\package\cf72f4f3a71ff1fe5381d0f0ebe068c867c28045\include\spdlog/details/circular_q.h(1,1): warning C4819: The file contains a character that cannot be represented in the current code page (949). Save the file in Unicode format to prevent data loss [C:\lab\cpp_starter_project\build\src\intro.vcxproj]
[build]   catch_main.vcxproj -> C:\lab\cpp_starter_project\build\lib\catch_main.lib
[build]   Building Custom Rule C:/lab/cpp_starter_project/test/CMakeLists.txt
[build]   Building Custom Rule C:/lab/cpp_starter_project/test/CMakeLists.txt
[build]   Building Custom Rule C:/lab/cpp_starter_project/test/CMakeLists.txt
[build]   tests.cpp
[build]   constexpr_tests.cpp
[build]   constexpr_tests.cpp
[...]

PR #96, opened 2020-10-08
Update multiple Conan dependencies.

ddalcino:
The `fmt` update is meant to address a bug reported in #81.

Note that `spdlog/1.8.0` is dependent on `fmt/7.0.3`.

Related to #65: specific dependency versions supersede #65, but this does not replace the move to a separate file. Unless these dependencies are moved to a separate file, the process of bumping these dependencies cannot be easily automated, and PRs like this one will always be necessary to keep packages up to date.

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -16,10 +16,10 @@ macro(run_conan)
   conan_cmake_run(
     REQUIRES
     ${CONAN_EXTRA_REQUIRES}
-    catch2/2.13.1
+    catch2/2.13.4
     docopt.cpp/0.6.3
-    fmt/7.0.3
-    spdlog/1.8.0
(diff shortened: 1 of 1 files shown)


ddalcino:
Please note: Although the Travis and Appveyor builds of this are working just fine, the Github Actions workflow for this is failing: https://github.com/ddalcino/cpp_starter_project/runs/1227289080

The error message is something to do with a conditional in fmt/core.h with duplicate branches, caused by this line in `src/main.cpp:44:41`:

    fmt::print("Hello, from {}\n", "{fmt}");

While the Travis builds use gcc-9, the Github Action build uses gcc-7. I cannot confirm, but I suspect that fmt/7.0.3 is expecting a more recent compiler.

PR #101, opened 2020-11-16
Set up code analysis flags for Visual Studio

aminya:
- This uses Cmake to set the flags required for code analysis inside Visual Studio.

- Updates the guide for the build step in Visual Studio


cmake/CodeAnalysis.cmake
```diff
@@ -0,0 +1,23 @@
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
+
+    # Read checks from the clang tidy file using PowerShell
+    execute_process(COMMAND powershell "Select-String -path ${CMAKE_CURRENT_SOURCE_DIR}/.clang-tidy \"Checks:[\\s\\t]*[`\"'](.*)[`\"']\" -AllMatches  | Foreach-Object {$_.Matches} | Foreach-Object {$_.Groups[1].Value}" OUTPUT_VARIABLE ClangTidyChecks)
```
aminya, line 4: VS seems to read the clang-tidy rules from the config file in the recent versions. So, this might not be needed anymore.

aminya: https://github.com/lefticus/cpp_starter_project/pull/101/commits/60c0a5e43ed4dda43628b57cd040a04317ffff1b

lefticus:
@aminya Interesting, this is features of CMake I had no idea existed.

cmake/Utils.cmake
```diff
@@ -0,0 +1,14 @@
...
+# - _dir root directory to start looking from
+function(get_all_targets _result _dir)
+    get_property(_subdirs DIRECTORY "${_dir}" PROPERTY SUBDIRECTORIES)
+    foreach(_subdir IN LISTS _subdirs)
+        get_all_targets(${_result} "${_subdir}")
+    endforeach()
+
+    get_directory_property(_sub_targets DIRECTORY "${_dir}" BUILDSYSTEM_TARGETS)
+    list(APPEND ${_result} ${_sub_targets})
+endfunction()
```
ddalcino, line 14: Part of the problem here is that CMake passes variables by value, not by reference. Any changes made within this function will remain local to the function and go out of scope on return. Unless you are using a macro (please don't call macros recursively), the only way to export results from a CMake function is to use `set(<variable_name> <value> PARENT_SCOPE)`.

Here is one approach to getting the results you intend:

```suggestion
function(get_all_targets _result _dir)
    get_property(_subdirs DIRECTORY "${_dir}" PROPERTY SUBDIRECTORIES)
    set(local_result "")
    foreach(_subdir IN LISTS _subdirs)
        get_all_targets(child_result "${_subdir}")
        list(APPEND local_result ${child_result})
    endforeach()

    get_directory_property(_sub_targets DIRECTORY "${_dir}" BUILDSYSTEM_TARGETS)
    foreach(_tgt IN LISTS _sub_targets)
        # We need to filter targets; many of them are not the targets we are looking for
        # In our case, the property TYPE is set to EXECUTABLE for the targets we want.
        # This may not be true in your case; you may need to modify this check to allow libraries as well.
        get_target_property(_target_type ${_tgt} TYPE)
        if(${_target_type} STREQUAL "EXECUTABLE")
            message(DEBUG "It's an executable target: ${_tgt}")
            list(APPEND local_result ${_tgt})
        else()
            message(DEBUG "Target ${_tgt} is a(n) ${_target_type}, not an executable")
        endif()
    endforeach()
    
    # If we don't export _result to the parent scope, none of our changes will be saved
    set(${_result} ${local_result} PARENT_SCOPE)
endfunction()
```

It is likely that we will want the property type check to use other criteria. TYPE == EXECUTABLE is convenient here, just because our targets are executables and not libraries; it would be better to allow libraries as well. However, we have "catch_main" as one of our targets, and that's a static library that we did not write. I don't think we need static analysis to run on "catch_main".

If you want to look into alternate property types we could use to filter out targets we don't want, [this stackoverflow answer](https://stackoverflow.com/questions/32183975/how-to-print-all-the-properties-of-a-target-in-cmake) has a few approaches to printing all of the properties available for a target.

ddalcino (changes requested):
You don't have to use my suggested code (perhaps you can write something better), but I feel very strongly that we need to generate a list of targets automatically.

PR #109, opened 2021-01-04
feat(sanitizers): Support non-interface targets

alireza-hosseini:
# What I did
In case the given target is not an interface, sanitizers shall be linked to it with `PRIVATE` visibility scope.

# How I tested
By using the `sanitizers.cmake` script in my personal project and enabling a few sanitizers for a lot of different interface and static libraries.


lefticus:
I think this looks like a potentially good idea, to make the script a bit more usable.

aminya (changes requested):
I think this PR is not needed anymore. You can link `project_options` as PRIVATE to your target. It does not need to be INTERFACE.

ddalcino:
I don’t think it makes sense to keep this PR open anymore. The file that it changes has moved to project-options. @alireza-hosseini, if you still feel that this project needs these changes, please open a new PR at https://github.com/cpp-best-practices/project_options.

PR #110, opened 2021-01-05
Remove duplicate header in template source file

Michael137:
Removes the duplicate "#include \<iostream\>". Unless this is for demonstrating warnings/static analysis. Though the default static analysis tools didn't pick it up.


ddalcino (approved):
Good catch; I don't see how this bug escaped notice for so long!

PR #113, opened 2021-01-17
Suppress max/min macros in windows.h

ddalcino:
Using MS Visual Studio, when you `#include <windows.h>`, some preprocessor macros are defined by default for `min()` and `max()`. As a result, `windows.h` interferes with the C++ STL, for example, `std::numeric_limits::min()` and `std::max()` in the algorithm header.

See [`minwindef.h`](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/shared/minwindef.h#LC190) for the offending code. This header is included by [`windef.h`](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/shared/windef.h#LC24), which is included by [`windows.h`](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/um/Windows.h#LC171). Thankfully, these definitions of `min` and `max` are surrounded by an `#ifndef NOMINMAX`, so we have a ready solution to the problem.

**Alternatives:** You could just not ever use the Visual Studio compiler; the `windows.h` file that ships with my MSys2 mingw-gcc does not have this problem. You could use some preprocessor #defines and/or #undefs in your otherwise pristine ISO-compliant C++ code. I prefer to use good cmake defaults that prevent this kind of thing from ever showing up. You shouldn't ever have to think about weird stuff like this.

**Potential for improvement:** I think that it might be worth moving this kind of code to a separate file in the cmake folder, perhaps called `cmake/Definitions.cmake`, but I'm not sure that's necessary.

**Background:** I just ran into this conflict between `windows.h` and `std::numeric_limits` in another project. When I googled the two terms together, I was so mad that so many other people complaining about this problem that I had to do something. Please, let's have sane defaults that make it easier to use the C++ STL, and not clobber the std namespace with legacy preprocessor macros!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,11 @@ if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
   endif()
 endif()
 
+if(MSVC)
+  # Prevent definition of `min` and `max` macros included by `windows.h`
+  target_compile_definitions(project_options INTERFACE NOMINMAX)
+endif()
+
 # Link this 'library' to use the warnings specified in CompilerWarnings.cmake
 add_library(project_warnings INTERFACE)
 


ddalcino:
I thought I should add some clarification as to why I think this PR belongs in the project, and what exactly this intends to fix.

Consider the following code:

```c++
#include <limits>
#include <windows.h>

constexpr auto get_amt() { return std::numeric_limits<int>::max(); }
```

When I compile this file with `mingw-gcc` or with `clang` on windows, without linking, compilation is successful. If I compile it with `cl.exe` (the MS Visual Studio compiler), compilation fails and I get this error:

```
limits.cpp(4): warning C4003: not enough arguments for function-like macro invocation 'max'
limits.cpp(4): error C2589: '(': illegal token on right side of '::'
limits.cpp(4): error C2062: type 'unknown-type' unexpected
limits.cpp(4): error C2059: syntax error: ')'
```

If I remove the `#include <windows.h>` line, `cl.exe` compiles the file correctly. This is because `windows.h` defines `min` and `max` as macros, and the preprocessor has no choice but to substitute the word `max` for that macro before evaluating my code.

This is not due to a bug in `cl.exe` or in `windows.h`. This is due to a design choice made decades ago by Microsoft, before `std::max` existed, and Microsoft cannot reverse that choice without breaking people's code.

The cleanest fix to this problem, other than abandoning MSVC (not an option for many of us), is to add the definition `NOMINMAX` to your project. This definition prevents the `min` and `max` macros from being defined. My solution puts that definition in a `CMakeLists.txt` file, upstream of any code where it would be needed. It's so far out of the way that, unless you go looking for it, you'll never even know this was a problem: you can just get to work, using `std::max` and `std::numeric_limits::max` without any problem.

Maybe this fix doesn't belong in this project. Maybe the text `NOMINMAX` has some meaning in some other library I don't know about, and by including this definition I am breaking someone else's project, though I doubt it. If so, I will happily close this PR unmerged.

To my understanding, the purpose of this project is to make it easier to write clean C++ that follows best-practices as much as possible. Once in a while, some of us have to get into the weeds with some platform-specific Windows code; when that happens, the `NOMINMAX` definition will be necessary in order to use some features of the STL. If you're not familiar with all the pitfalls involved with `windows.h`, you won't know what's happening when you see the error message listed above. This PR will prevent people from dealing with that headache, and make it more natural to write good C++.

PR #116, opened 2021-02-11
Require debug info for fuzzing

avitase:
Ensures that assertions, e.g., via `assert(...)` are processed and are not removed during compile-time optimizations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,8 +82,13 @@ if(ENABLE_TESTING)
 endif()
 
 if(ENABLE_FUZZING)
-  message("Building Fuzz Tests, using fuzzing sanitizer https://www.llvm.org/docs/LibFuzzer.html")
-  add_subdirectory(fuzz_test)
+  # Ensure that assertions are still checked
+  if (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+    message("Building Fuzz Tests, using fuzzing sanitizer https://www.llvm.org/docs/LibFuzzer.html")
+    add_subdirectory(fuzz_test)
+  else ()
+    message(FATAL_ERROR "Building and running fuzz tests is only supported in Debug configurations.")
+  endif ()
 endif()
 
 add_subdirectory(src)


CMakeLists.txt
```diff
@@ -82,8 +82,13 @@ if(ENABLE_TESTING)
 endif()
 
 if(ENABLE_FUZZING)
-  message("Building Fuzz Tests, using fuzzing sanitizer https://www.llvm.org/docs/LibFuzzer.html")
-  add_subdirectory(fuzz_test)
+  # Ensure that assertions are still checked
+  if (CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
```
ddalcino, line 86: It's my understanding that `RelWithDebInfo` is the same as `Release` but with debug symbols included. If that's true, then `RelWithDebInfo` will skip assertions as well.

If I were you, I would just check for `CMAKE_BUILD_TYPE STREQUAL "Debug"`.

Keep in mind that CMake allows you to create your own build types, and it's possible to create a build type that doesn't skip assertions that isn't called "Debug". On my machine, any build that defines the macro `NDEBUG` will skip assertions. If that behavior were standard C++ (I doubt that it is), maybe you could tie build failure to that macro definition in some way.

Even if checking for "Debug" build mode isn't a perfect solution to the problem, I think this is a really helpful addition to the project, and I hope it gets merged!

avitase: You are right. Maybe requiring a specific build type wasn't a good idea. What about this function:
```
constexpr void test_for_assert() noexcept {
    static_assert([] {
        int i = 0;
        assert(i++ == 0);
        return i;
    }() == 1);
}
```
https://godbolt.org/z/EPqbxn

This could be included exclusively for a Fuzz testing TU (after adding some `NOLINT`s). What do you think?

avitase: You are right. Maybe requiring a build type was not a good idea. What about this helper function:
```
constexpr void test_for_assert() noexcept {
    constexpr auto side_effect = 1;
    static_assert([] {
        int i = 0;
        assert(i++ == 0); // NOLINT
        return i;
    }() == side_effect);
}
```
https://godbolt.org/z/hjd3zM

This could be included exclusively for Fuzz testing TUs. The downside is that, AFAIK, this only works for C++17 or newer... What do you think?

ddalcino: That looks like well-defined C++17 code to me. It should not be a problem to use C++17 features; according to `CMakeLists.txt`, C++17 is required by this project anyway. I think this would work.

Maybe you could add something to the error message, like "Please use a debug build for fuzz testing", so it's easier for users to understand what went wrong?

avitase: I created a new pull request: https://github.com/lefticus/cpp_starter_project/pull/119

PR #119, opened 2021-02-24
Add assertions to fuzz test

avitase:
Add an example for using assertions within fuzz tests. This might be helpful since there is the caveat that when using assertions during fuzzing, one has to ensure that those assertions are not disabled, e.g., via `NDEBUG`. For larger projects one should factor out this testing for assertions but there are multiple ways to do this so inlining might be a good option for the sake of this exemplary file.


ddalcino:
That looks like a lot of extra overhead for anyone trying to read `fuzz_test/fuzz_tester.cpp`. The `test_for_assert()` code really should be in its own function, whose only purpose is to cause a compilation failure when assertions are disabled.

I was thinking that you were going to add this code to a header file, and include that file in every fuzz test .cpp file. I prefer that solution to this, since it's easier to extend to large projects.

Another thought: You could put the `test_for_assert()` function in its own .cpp file, and then link that file into every fuzz target from CMakeLists.txt. Users would never need to include the header file in any fuzz test .cpp file, and the `test_for_assert()` function name would never pollute the symbol table. You would never have to think about it; you could just trust that the compiler won't produce a fuzz_test binary with assertions disabled.

On the other hand, I really don't think it's so bad to just check that `CMAKE_BUILD_TYPE STREQUAL "Debug"`, as in #116; that would make this whole thing a lot simpler.

Perhaps we could get an opinion from @lefticus?

lefticus:
I have never seen a need to have an assertion inside my fuzz test. The point of the fuzz test is to simply run until an ASAN or UBSAN event is triggered.

Fuzz test, logically speaking, can never test for correctness. (because you don't know the correct answer to every possible input)

ddalcino:
I was supportive of this idea when it was in #116; maybe I should not have been.

I think that if your program being tested contains meaningful assertions, you would want your fuzz test to trip them. You want to know if it’s at all possible that your assertions could be false. If you’re using a release build, the assertions get stripped out and you will lose that information.

On the other hand, a release build on -O3 module could have a completely different set of branch points, and maybe you’d want to fuzz that too. This PR would prevent you from doing so.

ddalcino:
Ok, it has been a whole year since we have heard anything from @avitase on this. We have plenty of reasons listed above not to use something like this, so I will close it. @avitase, if you disagree with this decision, or you have anything further to add, please let us know and we will reopen it for you.

PR #121, opened 2021-02-26
Add Dockerfile

### LINKED ISSUE #120: Add docker support
Adding docker support would make the developer environment expected by this starter project portable. 

Setting up all the various executables expected by the cmake file is a little bit of a hinderance to productivity when setting up.

ddalcino:
This adds a Docker image with all of this project's required and optional dependencies preinstalled. This will make it easier for users to use all the tools this project offers.

~~This PR doesn't address the GUI examples; I am having a hard time getting the IMGUI and SDL examples to build from Docker.~~ The GUI examples build properly using a script.

Fixes #120.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,64 @@
+FROM ubuntu:bionic
+
+# Set up dependencies
+ENV GCC_VER "10"
+ENV LLVM_VER "11"
+ENV CC gcc-${GCC_VER}
+ENV CXX g++-${GCC_VER}
+
+# Add ability to install PPAs
+RUN apt-get update -qq
+RUN apt-get -y install software-properties-common wget git
+
+# Add repo for gcc-10
+RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test
+# Add repo for clang-11
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+RUN add-apt-repository -y "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-${LLVM_VER} main"
+# Add repo for current cmake/ccmake, from Kitware
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null \
+    | gpg --dearmor - | tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
+RUN apt-get update -qq
+
+# Install all the packages
+RUN apt-get install -y python3 python3-pip doxygen graphviz ccache cppcheck \
+    gcc-${GCC_VER} g++-${GCC_VER} \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
+RUN apt-get update -qq
+
+# Install all the packages
+RUN apt-get install -y python3 python3-pip doxygen graphviz ccache cppcheck \
+    gcc-${GCC_VER} g++-${GCC_VER} \
+    clang-${LLVM_VER} lldb-${LLVM_VER} lld-${LLVM_VER} clangd-${LLVM_VER} \
+    llvm-${LLVM_VER}-dev libclang-${LLVM_VER}-dev clang-tidy-${LLVM_VER} \
+    cmake cmake-curses-gui \
+    neovim emacs
```
ddalcino, line 30: Maybe add `nano`? Do people use `nano`? 

Does anyone care that it's `neovim` and not `vi`/`vim`?

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# `You need to copy the Clang include directory to the expected location before
+#  running (similarly, use include-what-you-use -print-resource-dir to learn
+#  exactly where IWYU wants the headers).`
+RUN mkdir -p $(include-what-you-use -print-resource-dir 2>/dev/null)
+RUN ln -s $(readlink -f /usr/lib/clang/${LLVM_VER}/include) \
+    $(include-what-you-use -print-resource-dir 2>/dev/null)/include
+
+# Include project
+ADD . /starter_project
+WORKDIR /starter_project
```
ddalcino, line 62: I'm not sure it's worth it to put a copy of the project in here. This copy will only serve to confuse users who want to mount their local source directory and work with the Docker image interactively.

Dockerfile
```diff
@@ -0,0 +1,64 @@
+FROM ubuntu:bionic
+
+# Set up dependencies
+ENV GCC_VER "10"
+ENV LLVM_VER "11"
+ENV CC gcc-${GCC_VER}
+ENV CXX g++-${GCC_VER}
```
ddalcino, line 7: Perhaps `CC` and `CXX` should be `ARG` and not `ENV`, so they can be passed as arguments to `docker build`?

ddalcino:
This is a whole lot of stuff to install/build, and it takes quite a while to build the image. I think this container might be worth posting to Dockerhub so our users don't have to rebuild the whole thing every time they use it. I intend to do that after we have a version of this that other people actually want to use.

ddalcino:
Also, there's a real issue with building some of the GUI examples in Docker. I can't get the IMGUI example to build at all, because the Conan recipe chokes on a `sudo apt-get update` command. `sudo` doesn't work here, because Docker logs you in as root by default. I'm not a Docker expert; maybe someone has a suggestion here?

ddalcino:
Update: All the GUI examples build successfully now, when you build them using the `build_examples.sh` script.

PR #122, opened 2021-02-27
Update Conan.cmake to fix #40

### LINKED ISSUE #40: conan doesn't recieve CC/CXX variables from CLion
Testcase:
* In CLion, define a profile with an alternative compiler (for me, it's clang-10)
* Clear the conan cache
* Have CLion regenerate the project

Expected result
* This should complete without errors, and allow the project to be built
   * All conan packages should be build

Actual result
* CLion calls something along the lines of
`/home/kees-jan/opt/clion-2019.3.5/bin/cmake/linux/bin/cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=/usr/bin/clang-10 -DCMAKE_CXX_COMPILER=/usr/bin/clang++-10 -G "CodeBlocks - Unix Makefiles" /home/kees-jan/src/cpp_starter_project`
* This triggers conan to build the dependencies
* That fails (among others in docopt) with
```
CMake Error at conanbuildinfo.cmake:553 (message):
  Incorrect 'clang', is not the one detected by CMake: 'GNU'
  conanbuildinfo.cmake:136 (conan_check_compiler)
  CMakeLists.txt:5 (CONAN_BASIC_SETUP)
```

Analysis
* Conan expects the `CC` and `CXX` variables to be set, in addition to the compiler settings, and they are not.
* I'm not sure whether to blame CLion, CMake, Conan, or this starter project for that shortcoming

Domkeykong:
This fixes #40 when building using an alternate compiler like `clang`.

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -22,6 +22,9 @@ macro(run_conan)
     spdlog/1.5.0
     OPTIONS
     ${CONAN_EXTRA_OPTIONS}
+    ENV 
+    "CC=${CMAKE_C_COMPILER}"
+    "CXX=${CMAKE_CXX_COMPILER}"
     BASIC_SETUP
     CMAKE_TARGETS # individual targets to link to
     BUILD


LtdSauce:
I can confirm that i still experienced the issue that this PR tries to solve with latest main branch. (Although i wasn't asked)

c60cb859:
> I can confirm that i still experienced the issue that this PR tries to solve with latest main branch. (Although i wasn't asked)

+1 from me. Running on Linux

ddalcino:
> > I can confirm that i still experienced the issue that this PR tries to solve with latest main branch. (Although i wasn't asked)
> 
> +1 from me. Running on Linux

Me too: see https://github.com/cpp-best-practices/cpp_starter_project/pull/175#discussion_r758840060. This PR resolved the issue for me. I must note that this was a very specific situation: [CLion + CMakePresets + Linux + clang}; if I change any one of these variables to anything else, I do not experience the problem.

ddalcino (changes requested):
Please merge upstream/main into this branch, and resolve the merge conflict. Note that `conan_cmake_run` has turned into `conan_cmake_install`, but the `ENV` argument that this PR adds is still a valid argument for `conan_cmake_install`.

ddalcino:
When #169 is merged, this PR will have less relevance here because the Conan.cmake file will be moved into [ProjectOptions](https://github.com/aminya/ProjectOptions).

@aminya, I think this change is worth adding to ProjectOptions. Several of our users have found this change useful, including myself (see above). I don't think that there is any downside to adding this: there should never be a situation where you would want Conan to use a different compiler than the one used by the project. What do you think?

You could argue that this change is only relevant when the user has failed to set CC and CXX manually, and that not setting these environment variables yourself is "user error". I prefer to forgive users for that "mistake", if it is one, particularly since this change won't hurt anything.

This is how I would expect Conan.cmake in ProjectOptions to be changed:
```diff
    conan_cmake_install(
      PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR}
      BUILD missing
            # Pass compile-time configured options into conan
      OPTIONS ${ProjectOptions_CONAN_OPTIONS}
+     ENV "CC=${CMAKE_C_COMPILER}" "CXX=${CMAKE_CXX_COMPILER}"
      SETTINGS ${settings})
```
I can file a PR for this myself, but I want to give @Domkeykong a chance to do it; it's @Domkeykong's PR, not mine.

aminya:
This is included in ProjectOptions v0.11.2
https://github.com/aminya/ProjectOptions

PR #126, opened 2021-03-02
Add conanfile.py

ddalcino:
This moves all hardcoded dependencies out of CMake code and into conanfile.py. Python code is responsible for conditionally adding dependencies, based on options passed to CMake.

This puts all the dependencies in one place, and makes the conditional dependency additions simpler to understand. According to Conan documentation, this is the recommended way to handle conditional dependencies.

Supersedes #124 and #65


aminya (approved):

lefticus (approved):
Approving due to comments discussed in #147

lefticus:
@ddalcino @aminya OK, I blocked myself out of this one, but forcing 2 code reviews for each PR, including for admins. We'll see how it goes. Also note that I moved the base to `main` as I've made that the new default branch.

PR #128, opened 2021-03-09
Add cmake helpers to cmake module path

HackingPheasant:
Overall this is more of a just Quality of Life improvement as it just removes visual cruft when including each CMake helper

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,12 @@ cmake_minimum_required(VERSION 3.15)
 
 # Set the project name to your project name, my project isn't very descriptive
 project(myproject CXX)
-include(cmake/StandardProjectSettings.cmake)
-include(cmake/PreventInSourceBuilds.cmake)
+
+# Add cmake helpers to cmake module path
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
+
+include(StandardProjectSettings)
+include(PreventInSourceBuilds)
 
 # Link this 'library' to set the c++ standard / compile-time options requested
 add_library(project_options INTERFACE)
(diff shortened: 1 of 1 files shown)


lefticus:
@ddalcino @aminya do you have an opinion on this one?

aminya (changes requested):
I would prefer the current behavior since the path to the CMake files is exactly as they are on the file system. If someone doesn't know what `CMAKE_MODULE_PATH` is, they might be confused for some time.

ddalcino:
> @ddalcino @aminya do you have an opinion on this one?

I really liked this PR, and I like to do this in my own projects. I forgot to give it an approving review ...

> I would prefer the current behavior since the path to the CMake files is exactly as they are on the file system. If someone doesn't know what CMAKE_MODULE_PATH is, they might be confused for some time.

... but that's a really good point, and I think @aminya is right.

lefticus:
I'll admit I did not actually know what it was doing at first either, and I've been using CMake for awhile now. I'm also inclined to the simple direct paths, as @aminya says.

Thank you @HackingPheasant , we appreciate your submission, but will not be merging it right now, as we think it could cause confusion for new users of CMake, and that outweighs the benefits.

HackingPheasant:
That's all good. I found it useful when I needed to include a cmake file in a CMakeLists.txt that was a few directories deep, meaning I didn't need to traverse a few directories and keep that path updated and correct if I ever did move stuff around.

And just to leave this info documented here if you ever change your mind you can have multiple paths listed like so (take from my personal project)

```
set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake;${PROJECT_SOURCE_DIR}/cmake/3rd-party" ${CMAKE_MODULE_PATH})
```

ddalcino:
> ```
> set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake;${PROJECT_SOURCE_DIR}/cmake/3rd-party" ${CMAKE_MODULE_PATH})
> ```

Personally, I'd like to see this line of code in the top level CMakeLists.txt file, commented out, with comments describing what it does and how it can be used. Users who want to use it (like me) could simply uncomment and start using it; others can just ignore it.

Issue #135, opened 2021-04-19
Align cmake-conan support using the last guidelines


salvioli:
The cmake script for Conan is outdated and is not aligned to the suggested use of the https://github.com/conan-io/cmake-conan wrapper. See the linked repository readme for last suggested way of using the wrapper.

PR #136, opened 2021-04-20
Update Conan.cmake to v0.16.1

wirew0rm:
Release notes since v0.15:
- [v0.16.0](https://github.com/conan-io/cmake-conan/releases/tag/v0.16)
- [v0.16.1](https://github.com/conan-io/cmake-conan/releases/tag/v0.16.1)

Also chage snippet according to conan cmake's [README](https://github.com/conan-io/cmake-conan/blob/develop/README.md).
Main differences:
 - directly link to github's cdn
 - enable verification of the TLS connection and the file hash.

fixes: #135

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -2,7 +2,10 @@ macro(run_conan)
   # Download automatically, you can also just copy the conan.cmake file
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
     message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
-    file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.15/conan.cmake" "${CMAKE_BINARY_DIR}/conan.cmake")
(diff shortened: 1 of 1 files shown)


prudhomm:
It actually fixes an issue with a macro parse_argument redefined which is quite nasty to find

ddalcino:
This change is definitely worth making. However, I don't think it's a real fix for #135. The conan_cmake docs state that conan_cmake_run "will be deprecated in the near future", and should be replaced with conan_cmake_configure, conan_cmake_autodetect, and conan_cmake_install, in that sequence.

wirew0rm:
Thanks for the feedback. I'll take another look and try to update the PR.

ddalcino:
I have some commits that may help you at https://github.com/ddalcino/cpp_starter_project/tree/fix_conan_cmake_usage; feel free to pull my commits or otherwise borrow code from them. My branch passes the CI tests, but I'm not sure that everything is set up right, and I may have broken the GUI examples. The IMGUI example looks ok, but the SDL2 example fails to build on my system right now.

Several things need to be changed after you remove `conan_cmake_run`, because `CONAN_PKG` doesn't exist anymore. I think you're meant to use `find_package`, and then `target_link_libraries` the same way you would if you weren't using Conan.

PR #141, opened 2021-06-10
FreeBSD 12.1 (default on Travis)

JeanVEGA:
Travis supports a FreeBSD 12.1 image. It's quite dated, but still useful for testing builds. [Successful run](https://travis-ci.com/github/JeanVEGA/cpp_starter_project/jobs/512657432).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,15 @@ jobs:
       env:
         - MATRIX_EVAL="CC=clang && CXX=clang++"
       addons: { apt: { packages: ['doxygen', 'python3-pip'] } }
+    - os: freebsd
+      compiler: clang
+      env:
+        - MATRIX_EVAL="CC=clang && CXX=clang++"
+      addons:
+        pkg:
+            - py37-pip
+            - conan
+
 
 
 before_script:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,10 @@ if(ENABLE_PCH)
     <utility>)
 endif()
 
+# spdlog async wants pthread_create
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+
 # Set up some extra Conan dependencies based on our needs before loading Conan
 set(CONAN_EXTRA_REQUIRES "")
 set(CONAN_EXTRA_OPTIONS "")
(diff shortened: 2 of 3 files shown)


ddalcino:
This looks potentially very useful, but I have some questions about the changes to `CMakeLists.txt` and `src/CMakeLists.txt`. 

Was it necessary to add support for spdlog async to get the project to compile and run on FreeBSD? To my understanding, the project is set up to use synchronous logging, so none of the changes to the CMakeLists.txt files should be necessary. 

I'm not trying to say that these changes are unappreciated; it's definitely nice to see a good example of how to turn on async logging in this project. I just want to know why they are part of a PR that claims to be focused on adding a FreeBSD build to Travis.

JeanVEGA:
Thanks for looking into this! Yes the change was necessary to make the
project build on FreeBSD. Without Threads the build fails
https://travis-ci.com/github/JeanVEGA/cpp_starter_project/jobs/512657432

> ld: error: undefined symbol: pthread_create

Should I look into explicitly disabling asynchronous logging support?

ddalcino:
Well, it's @lefticus's project; only he can tell you what he wants in the project.

If it were up to me, I would conditionally enable async logging with a CMake flag, and then turn on that flag in `.travis.yml` for the FreeBSD builds. That way, async logging would be available on all platforms, but not required, and turned off by default. Perhaps a note in the Readme could guide new users to turn the option on when compiling on FreeBSD.

Alternately, you could think about using CMake to detect the OS and turning the flag on when the OS is FreeBSD. I'm hesitant to recommend this though.

JeanVEGA:
Yes I noticed Travis has not been playing ball with open source projects recently. Also its support for FreeBSD is less than stellar and they do not seem to be [willing to change it](https://travis-ci.community/t/freebsd-vms-support/2125/3).

I've seen people run [*BSD VM on macos](https://github.com/marketplace/actions/freebsd-vm).

ddalcino:
> Yes I noticed Travis has not been playing ball with open source projects recently. Also its support for FreeBSD is less than stellar and they do not seem to be [willing to change it](https://travis-ci.community/t/freebsd-vms-support/2125/3).
> 
> I've seen people run [*BSD VM on macos](https://github.com/marketplace/actions/freebsd-vm).

Ok, good; so it looks like we're on the same page. In that case, I am going to close this PR.

I'd like to encourage you to open a new PR that uses the BSD VM on MacOS you've mentioned; I think that could be useful.

PR #146, opened 2021-07-23
warn about need for instrumented libs with MSan when using clang

arnemertz:
when clang is invoked with -fsanitize=memory, all libraries, including the standard library has to be built with that instrumentation to avoid false positives. This PR emits a warning for that case, so people do not have to google for the reason for failing CMake builds

https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo

### DIFF
--- a/cmake/Sanitizers.cmake
+++ b/cmake/Sanitizers.cmake
@@ -36,6 +36,7 @@ function(enable_sanitizers project_name)
 
     option(ENABLE_SANITIZER_MEMORY "Enable memory sanitizer" FALSE)
     if(ENABLE_SANITIZER_MEMORY AND CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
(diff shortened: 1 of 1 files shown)


aminya (approved):
Yeah, this is an important catch that I found out when wasting a lot of time trying to find the reason for the errors.

cmake/Sanitizers.cmake
```diff
@@ -36,6 +36,7 @@ function(enable_sanitizers project_name)
 
     option(ENABLE_SANITIZER_MEMORY "Enable memory sanitizer" FALSE)
     if(ENABLE_SANITIZER_MEMORY AND CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+      message(WARNING "Memory sanitizer reports false positives when running against libraries that are not MSan-instrumented (including libc++)")
```
aminya, line 39:
```suggestion
      message(WARNING "Memory sanitizer requires all the code (including libc++) to be MSan-instrumented otherwise it reports false positives")
```

PR #150, opened 2021-08-24
Update cmake conan

lefticus:
Upgrade uses of cmake-conan integration.

### DIFF
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -1,33 +1,38 @@
 macro(run_conan)
   # Download automatically, you can also just copy the conan.cmake file
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
-    message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
-    file(DOWNLOAD "https://github.com/conan-io/cmake-conan/raw/v0.16.1/conan.cmake" "${CMAKE_BINARY_DIR}/conan.cmake")
+    message(
+      STATUS
+        "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    file(
+      DOWNLOAD
+      "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
+      "${CMAKE_BINARY_DIR}/conan.cmake"
+      EXPECTED_HASH
+        SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
+      TLS_VERIFY ON)
   endif()
 
-  set (ENV{CONAN_REVISIONS_ENABLED} 1)
+  set(ENV{CONAN_REVISIONS_ENABLED} 1)
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
+  list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
+
   include(${CMAKE_BINARY_DIR}/conan.cmake)
 
-  conan_add_remote(
-    NAME
-    conan-center
-    URL
-    https://api.bintray.com/conan/conan/conan-center)
+  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
+  conan_add_remote(NAME cci URL https://center.conan.io INDEX 0)
 
-  conan_add_remote(
-    NAME
-    bincrafters
-    URL
-    https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
+  conan_cmake_autodetect(settings)
 
-  conan_cmake_run(
-    CONANFILE
-    conanfile.py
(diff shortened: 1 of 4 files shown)


lefticus:
Finally all builds looking green again

aminya (approved):
I was able to build the project successfully

cmake/Conan.cmake
```diff
@@ -1,33 +1,38 @@
...
-    NAME
-    bincrafters
-    URL
-    https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
+  conan_cmake_autodetect(settings)
 
-  conan_cmake_run(
-    CONANFILE
-    conanfile.py
+  conan_cmake_install(
```
ddalcino, line 27: I would like to add comments that make it abundantly clear what is happening here, and that `conan_cmake_install` is automatically detecting `conanfile.py` and running it. The cmake-conan readme makes no mention of this; I had to go digging in the issues at https://github.com/conan-io/cmake-conan/issues/331 to find this out. I'm sure that anyone reading this code alongside the cmake-conan readme will wonder why we skipped the call to `conan_cmake_configure`, which the readme recommends doing. 

For the record, calling `conan_cmake_configure` autogenerates a Conanfile.txt; there's no need to do that if you already have a Conanfile.py.

lefticus: Right, that's a good point. I considered adding a comment for the directory used in conan_cmake_install. I'll do that now.

src/CMakeLists.txt
```diff
@@ -41,12 +41,16 @@ if(CPP_STARTER_USE_NANA)
...
 target_link_libraries(
   intro
   PRIVATE project_options
           project_warnings
-          CONAN_PKG::docopt.cpp
-          CONAN_PKG::fmt
-          CONAN_PKG::spdlog)
+          docopt::docopt
+          fmt::fmt
+          spdlog::spdlog)
```
ddalcino, line 56: This looks good. Keep in mind, we need to give the same treatment to the CMakeLists.txt files in the imgui and sdl examples as well. I did this in another branch that tried to tackle this issue here: https://github.com/ddalcino/cpp_starter_project/commit/edd75c7938b5aa9e94b52e7f3ceb10ad26455c30. I did have this working at one point, but I'm not certain it still works now.

lefticus: Oh, sorry for stepping on work you had already done!

lefticus:
We might also want to address the Windows multi-config, for working from inside of VS

https://github.com/conan-io/cmake-conan#using-conan_cmake_autodetect-and-conan_cmake_install-with-multi-configuration-generators

cmake/Conan.cmake
ddalcino: Much improved!

src/CMakeLists.txt
ddalcino: No problem! Your attempt is better anyway, since it's up to date on the whole Conanfile situation. I never made a PR out of this, so it's my own fault. I mentioned it in the discussion about #136 and then forgot about it.

ddalcino:
> We might also want to address the Windows multi-config, for working from inside of VS
> 
> https://github.com/conan-io/cmake-conan#using-conan_cmake_autodetect-and-conan_cmake_install-with-multi-configuration-generators

Good point.

BTW, does anyone know what happened to the Travis-CI builds? I don't see them in the 'checks' section anymore. Is it because they changed how they provision compute time for open source projects, or because they changed their url from travis-ci.org to travis-ci.com?

cmake/Conan.cmake
```diff
@@ -1,33 +1,44 @@
...
-    https://api.bintray.com/conan/conan/conan-center)
+  # Add (or remove) remotes as needed
+  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
+  conan_add_remote(NAME cci URL https://center.conan.io INDEX 0)
 
-  conan_add_remote(
-    NAME
-    bincrafters
-    URL
-    https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
```
ddalcino, line 21: I am unclear as to why this remote was removed. I understand the concept of "Add or remove remotes as needed", but this remote is absolutely required if you want to get the SDL example running. Perhaps this was done unintentionally?

ddalcino: Additionally, in case anyone else is trying to get the SDL example to build, I'd like to share that you will need to run it using a Python before 3.9. On Linux, SDL's dependency tree includes mesa, and some of the Python code that builds mesa relies on some crufty XML parser code that was deprecated many years ago, and was removed in Python 3.9.

lefticus:
> 
> BTW, does anyone know what happened to the Travis-CI builds? I don't see them in the 'checks' section anymore. Is it because they changed how they provision compute time for open source projects, or because they changed their url from travis-ci.org to travis-ci.com?

I had kind of given up on travis and forgot about it entirely.

Related? 
image

PR #151, opened 2021-08-24
Add a basic gitignore

aminya:
This pull request adds a basic gitignore. I deliberately made it simple so that it only covers the common ground. 

A simplified version of #59


ddalcino (approved):
I agree that this is worth having.

I'm a little wary of merging this without @lefticus' explicit approval. #59 was another instance where both @aminya and I thought that some version of a .gitignore should be merged, but @lefticus disagreed.

PR #152, opened 2021-08-25
Allow users to turn off Conan

ddalcino:
This PR allows users to opt out of using Conan for dependency management. It can be very challenging to get Conan to successfully build some dependencies (that means you, SDL2!), and I think there's some value in allowing users to turn Conan on and off at will, if they think it will be easier to satisfy those dependencies in some other way.

This change depends on some changes made by #150; currently, `src/CMakeLists.txt` depends directly on the use of Conan to satisfy package management, but #150 will separate the build system from Conan enough that the dependencies can be fulfilled in some other way.

I would like to request that this PR not be merged until after #150 is merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,8 +60,11 @@ if(ENABLE_PCH)
     <utility>)
 endif()
 
-include(cmake/Conan.cmake)
-run_conan()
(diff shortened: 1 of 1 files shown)


aminya (approved):

lefticus:
Agreed, good idea. I'm really happy with moving away from the `CONAN_PKG::` style.

PR #159, opened 2021-08-27
Run code format check in Github Actions

ddalcino:
This adds an extra Github Action to perform code format and lint checks on every push. Currently, it causes a failed check if CppCheck finds problems, or if clang-format would make changes to the code, or if cmake-lint disagrees with the style of your CMake code.

There are other ways to accomplish this task. I am concerned that cmake-lint might be too aggressive a tool for this purpose; it may be sufficient to just check if cmake-format would make changes.

Some projects choose to use clang-format and cmake-format in CI in a way that causes a bot to generate PRs to fix code formatting; that's another reasonable option.

The whole point of this PR is to add the `lint-check.yml` file; the changes in the other 9 files are simply a product of running clang-format and cmake-format so that the new checks pass.

**Edit:** This PR has been changed to remove the C++ static analysis workflow; that work can be saved for a future PR, possibly using https://github.com/aminya/setup-cpp when it supports IWYU.

### DIFF
--- a/.github/workflows/code-format-check.yml
+++ b/.github/workflows/code-format-check.yml
@@ -3,9 +3,6 @@ name: Code Format Check
 on:
   pull_request:
   push:
-    branches:
-      - main
-      - master
 
 defaults:
   run:
--- a/.github/workflows/code-format-check.yml
+++ b/.github/workflows/code-format-check.yml
@@ -1,0 +1,45 @@
+name: Code Format Check
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - master
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: 'ubuntu-latest'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install tools
+        run: |
+          pip install cmakelang
+
+      - name: Check cpp code formatting
+        run: |
+          git ls-files |\
+            grep -E '\.(c|cc|C|cii|cpp|cxx|c++|h|hii|hpp|hxx|h++)$' |\
+            xargs clang-format --dry-run --Werror
+
+      - name: Check CMake code formatting
+        run: |
+          git ls-files |\
+            grep -E '(\.cmake$|CMakeLists.txt)' |\
+            xargs cmake-format --check
+
+      - name: Lint CMake code
+        # This step is configured to print all the output that cmake-lint would
+        # generate, but it will not cause a build job failure.
+        # If you need this step to fail when cmake-lint reports issues,
+        # simply remove the `|| true` from the last line.
+        run: |
+          git ls-files |\
+            grep -E '(\.cmake$|CMakeLists.txt)' |\
+            xargs cmake-lint || true
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -1,15 +1,11 @@
 macro(run_conan)
   # Download automatically, you can also just copy the conan.cmake file
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
-    message(
-      STATUS
-        "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
     file(
-      DOWNLOAD
-      "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
+      DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
       "${CMAKE_BINARY_DIR}/conan.cmake"
-      EXPECTED_HASH
-        SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
+      EXPECTED_HASH SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
       TLS_VERIFY ON)
   endif()
 
@@ -19,11 +15,18 @@ macro(run_conan)
 
   include(${CMAKE_BINARY_DIR}/conan.cmake)
 
-  # Add (or remove) remotes as needed
-  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
-  conan_add_remote(NAME cci URL https://center.conan.io INDEX 0)
+  # Add (or remove) remotes as needed conan_add_remote(NAME conan-center URL https://conan.bintray.com)
   conan_add_remote(
-    NAME bincrafters URL
+    NAME
+    cci
+    URL
+    https://center.conan.io
+    INDEX
+    0)
+  conan_add_remote(
+    NAME
+    bincrafters
+    URL
     https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
 
   # For multi configuration generators, like VS and XCode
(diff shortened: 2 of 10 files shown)


.github/workflows/lint-check.yml
```diff
@@ -0,0 +1,48 @@
+name: Static Analysis
+
+on:
+  pull_request:
+  push:
+
```
ddalcino, line 6: This deserves comment, since it is different from `build_cmake.yml`, which runs on PRs and pushes to the main/master branch. This check runs on *all* pushes.

Personally, I think it's very useful to run both the build-and-test jobs and the lint-check jobs on every push, so I get immediate feedback on whatever I've done wrong. I too often forget, or am otherwise unable, to run all of these checks locally. I think it's a sane default to run all the checks on all pushes.

On the other hand, running these checks isn't free, and it makes sense not to waste CI minutes indiscriminately. I have blown through all of my free CI minutes for the month doing this. It might be right to restrict these checks to PRs and pushes to main/master, as in `build_cmake.yml`.

.github/workflows/lint-check.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Lint CMake code
+        # This step is configured to print all the output that cmake-lint would
+        # generate, but it will not cause a build job failure.
+        # If you need this step to fail when cmake-lint reports issues,
+        # simply remove the `|| true` from the last line.
+        run: |
+          git ls-files |\
+            grep -E '(\.cmake$|CMakeLists.txt)' |\
+            xargs cmake-lint || true
```
ddalcino, line 48: I'm allowing failing `cmake-lint` checks not to cause a build failure, because some of the things it checks are frankly ridiculous. For instance, I get this warning on Sanitizers.cmake:

```
cmake/Sanitizers.cmake:02: [R0912] Too many branches 13/12
```

Excuse me? I'm not allowed to use more than 12 `if` statements? What's the justification for that?

I am aware that `cmake-lint` is configurable with a yaml file, or with CLI arguments, but I haven't done anything with that yet. There is an option to increase the max number of allowed `if` statements.

If you want to see the output from `cmake-lint`, here's a recent run: https://github.com/cpp-best-practices/cpp_starter_project/runs/3446661973?check_suite_focus=true#step:7:1. I think a lot of what it recommends is useful advice, particularly if we want this project to be an example of excellent CMake code that others should emulate/copy.

ddalcino: I think this needs to be adjusted to use the static analyzers set up in CMake code, which already knows how they are meant to be run.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Cache
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-conan-modules
+        with:
+          path: |
+            ${{ env.CONAN_USER_HOME }}
+            ~/.cache/pip
+          key: ${{ runner.os }}-${{ env.BUILD_TYPE }}-${{ hashFiles('CMakeLists.txt') }}-${{ hashFiles('conanfile.py') }}
```
aminya, line 90: If you use the llvm setup action, this cache can also include the LLVM installation. See
https://github.com/kylemayes/install-llvm-action#example-usage-with-caching

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Build and test
+        run: |
+          ctest --build-and-test . ./build_dir \
+                --build-generator "Unix Makefiles" -j2 \
```
aminya, line 95: A separate build step is better than relying on the unix specific make files.
```suggestion
                --build-generator "Unix Makefiles" -j2 \
```

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      - name: Install python dependencies
+        run: |
+          python --version | grep -iE "python\s*3"  # fail if python 2 is default
+          pip    --version | grep -iE "python\s*3"
+          pip install cmake conan
+
+      - name: Install cppcheck
+        run: |
+          sudo apt-get update -qq
```
aminya, line 42: Probably not needed
```suggestion
```

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+          su && echo -e '\U0001F9B8' # 🦸
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - 2>/dev/null
+          add-apt-repository -y "deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic-${LLVM_VER} main"
+          apt-get update -qq
+          apt-get install -y --no-install-recommends \
+            clang-${LLVM_VER} lldb-${LLVM_VER} lld-${LLVM_VER} clangd-${LLVM_VER} \
+            llvm-${LLVM_VER}-dev libclang-${LLVM_VER}-dev clang-tidy-${LLVM_VER}
+          update-alternatives --install /usr/bin/clang-tidy clang-tidy $(which clang-tidy-${LLVM_VER}) 10
+          update-alternatives --install /usr/bin/clang   clang   $(which clang-${LLVM_VER}) 10
+          update-alternatives --install /usr/bin/clang++ clang++ $(which clang++-${LLVM_VER}) 10
```
aminya, line 65: Please use https://github.com/kylemayes/install-llvm-action instead. You can use it to easily install LLVM.

To setup libclang, you only need this:
https://github.com/kylemayes/install-llvm-action#linking-to-installed-libraries-linux

aminya (changes requested):

.github/workflows/static-analysis.yml
ddalcino: The real problem here is IWYU. That's the only reason to install any of this stuff. There is an [action](https://github.com/emilgedda/include-what-you-use-action) you can use to install it, but it locks you into using clang-9. It won't work if you need to use newer C++ features.

I think we want to be able to build and install IWYU with an arbitrary version of clang and cache the result of that process. IWYU takes much longer to build than it takes to install any of the clang packages.

If you're interested in taking over that job, I welcome the effort. I'm not going to have time to do it myself for a while.

aminya:
I have started developing the ever-wanted setup-cpp package. I think if we can make that package working, setting up cross-platform environments would be very easy.
https://github.com/aminya/setup-cpp

ddalcino:
> I have started developing the ever-wanted setup-cpp package. I think if we can make that package working, setting up cross-platform environments would be very easy.
> https://github.com/aminya/setup-cpp

Looks cool! I’ll check it out later, and if I have time I might send you a PR that installs IWYU. It’s not too hard to do on Mac or Linux, but Windows can be very challenging.

ddalcino:
I am considering removing the `.github/workflows/lint-check.yml` workflow entirely. If we can't run IWYU (yet), this file is mostly a duplicate of the existing `build_cmake.yml`; it would make more sense to add cppcheck and clang-tidy to that workflow.

Another option would be to restrict this PR to the code-formatting checks in `code-format-check.yml`, and add a lint check in a separate PR.

.github/workflows/lint-check.yml
ddalcino: File removed; this can be picked up in another PR

ddalcino:
Superseded by #181

lefticus:
Oh @ddalcino so sorry to have stepped on this PR. I got carried away after I implemented this in a client's code base.

ddalcino:
No problem. It would have taken a lot of effort to rewrite this PR to fit in with the setup-cpp and project_options changes, and honestly I don’t really have the time to do it right now.

PR #168, opened 2021-09-13
ci: use gcovr on windows

aminya:
A continuation of #165. [gcovr supports Windows](https://www.mankier.com/1/gcovr#Installation) too, and so I don't think we need to use a separate tool for Windows.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -56,19 +56,17 @@ jobs:
         pip3 install wheel setuptools gcovr
         pip3 install conan --upgrade
 
-    - name: Install / Configure dependencies
-      shell: bash
+    - name: Linux - Fix Conan Path
+      if: runner.os == 'Linux'
       run: |
-        if [ "$RUNNER_OS" == "Linux" ]; then
-          echo "Ubuntu - Fix Conan Path"
           sudo update-alternatives --install /usr/bin/conan conan /home/runner/.local/bin/conan 10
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -30,13 +31,8 @@ jobs:
         os: ['ubuntu-latest', 'windows-latest', 'macos-latest']
         include:
         - os: ubuntu-latest
-          ENABLE_COVERAGE: ON
         - os: windows-latest
-          # Windows works differently
-          ENABLE_COVERAGE: OFF
         - os: macos-latest
```
ddalcino, line 35: Now that `ENABLE_COVERAGE` is `ON` for all platforms, is this `include` block necessary?

aminya: No. We can remove it.

aminya:
In the gcovr tests, it does not say that it supports MSVC. But Clang and GCC are there.
https://www.mankier.com/1/gcovr#Installation

ddalcino:
> In the gcovr manual, it does not say that it supports MSVC. But Clang and GCC are there.
> https://www.mankier.com/1/gcovr#Installation

This looks really promising, if you can get reporting to work on Windows.

If you can't get gcovr to work with MSVC, maybe it will work with the Windows versions of clang or mingw.

aminya:
I am going to redo this in a new way.

Issue #147, opened 2021-08-02
Meta: add more repository maintainers


aminya:
I am opening this issue specifically because it seems that @lefticus is not able to review and merge all the pull requests, and so many of them are left open. One way to fix this is to add more maintainers to the repository.

27 PRs are open at the moment
https://github.com/lefticus/cpp_starter_project/pulls

ddalcino:
Agreed. It's a lot of work to maintain a popular repository like this, and @lefticus should not feel bad about not having enough time to do it. I think there are lots of people here that would be willing to help.

The last commits were merged more than 7 months ago. It's easy to get the impression that this project was abandoned, and I don't think that's the intention here.

lefticus:
That's a fair point. For this project it's really more a matter of forgetting to check and see that I have open PR's, and less about available time! I'll dig through this all shortly and come back to this issue.

lefticus:
@ddalcino and @aminya 

The two of you are the most active on here, and the obvious candidates for helping out with PRs.

You generally seem to agree with the same things I do, as well. My only comment is that I am opposed to the external conan configuration file, and want to keep things internal as conan_cmake integration.

What are your thoughts about this?

aminya:
I would be happy to be able to help! 

> You generally seem to agree with the same things I do, as well. My only comment is that I am opposed to the external conan configuration file, and want to keep things internal as conan_cmake integration.

If by an external Conan file you mean #126, I have to say that currently the dependencies are scattered all around the CMake files. Having a central place to manage all of those is much easier than remembering where a dependency is stored. 

However, I am also more biased over a simple Conan file like #65. Since it is a simple list of the dependencies and versions.

lefticus:
@aminya My observation in real-world open source (but commercially funded) projects is that if the user has to run `conan` as a separate tool (that you can forget to run, or can get out of date accidentally), then that is a non-starter.

So I want full cmake integration, so that builds are

```
<configure>
<build>
```

and there is no intermediate

```
<configure>
<run other tools (aka conan)
<build>
```

From the projects I've worked on, this is more important than you'd think, so I want to keep C++ Starter Project as simple to configure and build as possible, and leave the option of an external conanfile.py up to the user after forking/templating this project, but not a requirement to getting started.

This is why I've avoided merging any of the previous conanfile PR's.

ddalcino:
> The two of you are the most active on here, and the obvious candidates for helping out with PRs.

[...]

PR #169, opened 2021-09-22
feat: make the build system reusable and declarative + support Vcpkg + automatically setup MSVC environment 

### LINKED ISSUE #40: conan doesn't recieve CC/CXX variables from CLion

### LINKED ISSUE #125: Make the Cmake files a Cmake library
Whenever I want to use this project, I have to copy a bunch of boilerplate Cmake code to different places, which is rather inconvenient and error-prone. It would be nice if the code can become a Cmake library to solve this issue of copying.

By library I mean something similar to ixm:
https://ixm.one/

This is all I need to install `ixm`:
```cmake
include(FetchContent)
FetchContent_Declare(ixm URL https://get.ixm.one)
FetchContent_MakeAvailable(ixm)
```

### LINKED ISSUE #174: clang-tidy does checks using c++14 by default (not what specified for the compiler)
Trying to compile a sample project with Qt6 and clang-tidy enabled gives me the following error:
```
/usr/include/qt6/QtCore/qglobal.h:96:6: error: "Qt requires a C++17 compiler" [clang-diagnostic-error]
#    error "Qt requires a C++17 compiler"
```
because clang++ runs default with c++14 and not c++17 or with the setting from cmake.
Adding `-extra-arg=-std=c++17` in cmake/StaticAnalyzers.cmake in the clang-tidy part, fixes my problem.
I am not sure how to make it inherit the standard set for the compiler options.

aminya:
Fixes #125 
Fixes #174
Fixes #40
Fixes #122 

This pull request makes the build system reusable by:
- Making the Cmake files a Cmake library that is downloadable and reusable from multiple projects. 
   https://github.com/aminya/ProjectOptions/

- Removing the need for the copy-pasting 500 lines of CMake into different projects

This makes the build system declarative and reproducible by:
- Removing dynamic dependency injection
- Removing unnecessarily Cmake options that can be instead declared

All the options are exposed as function arguments and can be enabled/added by the user in a declarative way. Optionally there is a way to use global options.

It also improves upon many of the CMake files. For example, support for Vcpkg is added now, vcvarsall automatically runs, MSVC address sanitizer now works, etc.
See this for some of the improvements:
https://github.com/aminya/ProjectOptions/releases

See this for more information:
https://github.com/cpp-best-practices/cpp_starter_project/issues/125#issuecomment-924479679


conanfile.py
```diff
@@ -2,25 +2,14 @@
...
-    }
     name = "CppStarterProject"
     version = "0.1"
     requires = (
         "catch2/2.13.7",
         "docopt.cpp/0.6.2",
         "fmt/8.0.1",
         "spdlog/1.9.2",
+        # "imgui-sfml/2.1@bincrafters/stable"
+        # "sdl2/2.0.10@bincrafters/stable"
```
ddalcino, line 13: I know you're trying to make `cmakelib` more reusable, and that means not having variables like `CPP_STARTER_USE_IMGUI` and `CPP_STARTER_USE_SDL` in it, but there must be some other way to set options that will be passed directly to `conan_cmake_install`. Perhaps the consumer of `cmakelib` can set a `CONAN_OPTIONS` variable and call `cmakelib(... EXTRA_CONAN_OPTIONS ${CONAN_OPTIONS}`, and then `cmakelib()` can call `conan_cmake_install(... ${EXTRA_CONAN_OPTIONS})`?

Have you considered making this change as a separate PR?

If we are going this direction, I think it would make more sense to replace the whole `conanfile.py` with a `conanfile.txt`, and then users can uncomment `imgui` and `sdl2` as needed. As is, this change makes it impossible to install `imgui` and `sdl2` when you run the CMake configure step, and that could be important for testing the sdl and imgui subprojects in CI. This goes against the ["one step sets up everything" ideal](https://github.com/cpp-best-practices/cpp_starter_project/issues/147#issuecomment-904277057) that @lefticus was going for initially.

Personally, I don't have a problem with this. At best, the conan `sdl2` package takes forever to build and it usually doesn't work anyway. The situation with `imgui-sfml` is much better, but I think the overhead involved in building it is very high compared to just installing a prebuilt binary. In CI at least, I think it's a big improvement to use a package manager for this rather than build these dependencies with Conan.

ddalcino:
> See this for more information:
> [#125 (comment)](https://github.com/cpp-best-practices/cpp_starter_project/issues/125#issuecomment-924479679)

Regarding this comment, I think it would be useful to move cmakelib into the cpp-best-practices organization. I don't think I can approve this until that happens.

conanfile.py
aminya: Fixed in https://github.com/cpp-best-practices/cpp_starter_project/pull/169/commits/e0d227fd6952bf0f96083ad4365e751a2af4af80

aminya: I prefer that we remove all the unnecessary and fragile dynamism from the codebase. At the end of the day, people want to decide on one or more libraries and only use those. There is no need to add global mutable options that make the build fragile and unnecessarily dynamic.

ddalcino: In that case, please replace this file with a `conanfile.txt`. After these changes, there is no benefit to using a `conanfile.py`.

aminya: Fixed in https://github.com/cpp-best-practices/cpp_starter_project/pull/169/commits/e6b4af38ab46df8f840824b1f05b9fc3d6e7597b

aminya:
> > See this for more information:
> > [#125 (comment)](https://github.com/cpp-best-practices/cpp_starter_project/issues/125#issuecomment-924479679)
> 
> Regarding this comment, I think it would be useful to move the `cmakelib` project into the cpp-best-practices organization. I don't think it makes sense for me to approve this before that happens.

I am awaiting @lefticus to give me an invitation so I can transfer the repository. In the meantime, feel free to suggest a better name for `cmakelib`.
https://github.com/aminya/cmakelib/issues/5

build_examples.sh
```diff
@@ -1,133 +1,42 @@
...
-  mkdir -p "${BUILD_DIR}"
   # Note: Nana's headers trigger the -Wshadow warning, and cannot be compiled with -Werror.
   # Supposedly, this is fixed in the develop-1.8 branch, but I cannot confirm.
-  cmake -S "${PROJECT_DIRNAME}" -B "${BUILD_DIR}" -DCPP_STARTER_USE_NANA=ON -DWARNINGS_AS_ERRORS=FALSE
-  cmake --build "${BUILD_DIR}" -j
 }
 
-function build_qt() {
+function install_qt() {
   apt-get install -y --no-install-recommends qt5-default qtbase5-dev
```
aminya, line 30: Qt is available from Conan. Not sure it is good practice to use the system-wide dependency! 
https://conan.io/center/qt

CMakeLists.txt
```diff
@@ -2,85 +2,53 @@ cmake_minimum_required(VERSION 3.15)
...
+# Add cmakelib
+include(FetchContent)
+FetchContent_Declare(cmakelib URL https://github.com/aminya/cmakelib/archive/refs/heads/main.zip)
+FetchContent_MakeAvailable(cmakelib)
+include(${cmakelib_SOURCE_DIR}/Index.cmake)
+
+# Initialize cmakelib
+# uncomment the options to enable them
+cmakelib(
+    Enable_CACHE
```
ClausKlein, line 15: Why mixed case, all other options are upper case?

CMakeLists.txt
```diff
@@ -2,85 +2,53 @@ cmake_minimum_required(VERSION 3.15)
...
-include(cmake/StandardProjectSettings.cmake)
-include(cmake/PreventInSourceBuilds.cmake)
 
-# Link this 'library' to set the c++ standard / compile-time options requested
-add_library(project_options INTERFACE)
+# Add cmakelib
+include(FetchContent)
+FetchContent_Declare(cmakelib URL https://github.com/aminya/cmakelib/archive/refs/heads/main.zip)
+FetchContent_MakeAvailable(cmakelib)
+include(${cmakelib_SOURCE_DIR}/Index.cmake)
```
ClausKlein, line 10: Would it be an option to use `CPM.cmake` module to hide  `FetchContent`?
It works like this:
```cmake
# ---- Add dependencies via CPM ----
# see https://github.com/cpm-cmake/CPM.cmake for more info

include(cmake/CPM.cmake)

# PackageProject.cmake will be used to make our target installable
CPMAddPackage("gh:TheLartians/PackageProject.cmake@1.7.0")
# ...
packageProject(
  NAME ${PROJECT_NAME}
  VERSION ${PROJECT_VERSION}
  NAMESPACE ${PROJECT_NAME}
  BINARY_DIR ${PROJECT_BINARY_DIR}
  INCLUDE_DIR ${PROJECT_SOURCE_DIR}/include
  INCLUDE_DESTINATION ${INCLUDE_INSTALL_DIR}
  COMPATIBILITY SameMajorVersion
  # XXX DEPENDENCIES "fmt 7.1.3"
)
````

Comment (ClausKlein) [reply CMakeLists.txt:10]: for an usage example see this patch:
https://github.com/ClausKlein/cpp_starter_project/blob/3337f68032fbd0138edf91cfdbd09f7d9de59ae1/CMakeLists.txt#L29

File: CMakeLists.txt (hunk 3 above)
Comment (aminya) [reply CMakeLists.txt:15]:
```suggestion
    ENABLE_CACHE
```

CMakeLists.txt
aminya: I prefer that we use built-in libraries. All we want to do is to somehow download cmakelib. I will eventually add git tags, so the downloading will be more reliable.

I think CPM is useful for actual libraries.

CMakeLists.txt
```diff
@@ -2,85 +2,53 @@ cmake_minimum_required(VERSION 3.15)
...
+    # ENABLE_DOXYGEN
+    # ENABLE_IPO
+    # ENABLE_USER_LINKER
+    # ENABLE_BUILD_WITH_TIME_TRACE
+    # ENABLE_UNITY
+    # ENABLE_SANITIZER_ADDRESS
+    # ENABLE_SANITIZER_LEAK
+    # ENABLE_SANITIZER_UNDEFINED_BEHAVIOR
+    # ENABLE_SANITIZER_THREAD
+    # ENABLE_SANITIZER_MEMORY
```
ddalcino, line 32: I can see what you're doing here, and I think there's some value in changing some of these options into constants that cannot be changed without modifying CMake code.

However, I find it very useful to be able to turn many of these options on and off at will, without modifying any code. For instance, if you're using the sanitizers, the only sensible way to use them is by turning them on and off. Most (all?) of them are incompatible with one another, and if two are turned on at once, they will interfere with each other. If you're using them, they need to be turned on and off one at a time. It is true that you can just uncomment the options you need, whenever you need them, but for a lot of use cases, I think it's more practical to do it with a CMake option.

If you need a CI build that runs the sanitizers, how would you turn them on without affecting the other CI builds? How about a build specifically to run the static analyzers, or doxygen to build your documentation? If you need all of these features, are you just supposed to turn them all on at once? Some of these features make more sense to run in CI (ie doxygen, coverage) than locally on a developer's machine. It is not appropriate for me to have `ENABLE_COVERAGE` turned on while I am trying to debug some C++ code, and I don't like having to reach in and modify CMake files so I can get some work done.

I think it might make sense to have `ENABLE_CACHE`, `WARNINGS_AS_ERRORS`, and maybe `ENABLE_UNITY` to be hardcoded, but I can think of reasons for most of the others to remain as CMake configuration options. I am not sure about `ENABLE_PCH`, `ENABLE_UNITY` and `ENABLE_IPO`; this is likely due to my unfamiliarity with those tools and/or lack of imagination.

build_examples.sh
ddalcino: I don't think it's a good idea to just use Conan for all dependency management; Conan is sometimes problematic, and I think it's best to give people options when they need them.

Depending on what parts of Qt you need, you could be downloading 17MB or multiple gigabytes. The Conan package doesn't appear to give you any options; it appears to install ~600MB, and takes a very long time to do it.

Not to plug my own projects, but there's also [aqtinstall](https://github.com/miurahr/aqtinstall) for installing Qt via command line (it works on Win/Mac/Linux), and [a Github Action](https://github.com/jurplel/install-qt-action) based on it. You can use either aqt or apt to install the minimum 17MB version, and IMHO if that's all you need that's a whole lot better than what Conan offers.

This change turns `build_examples.sh` into a documentation file, rather than a program that can be executed. If that's all it is now, I would rather have it tell you that you have multiple ways to install Qt, depending on your preferences.

ddalcino (commented):
I like the direction that this is going, and I think that replacing most of this project with something like `cmakelib` could be really useful to a lot of people.

Given the changes you have made, it is no longer possible to build any of the GUI examples without altering CMake code. I think this is a problem. As long as those examples exist in the code base, I think we need to be able to build them from the command line. 

There are certainly ways to make this work. One option would be to move them out of the `src` folder entirely, and into a parallel `examples` folder. Each example could have its own top-level CMakeLists.txt and conanfile.txt files; you could run `cmake -S ./examples/fltk ...` to build the fltk example. I don't really like this solution, but it is more declarative than what we have now, and it's better than leaving the examples inoperable. Perhaps you can come up with another solution that you like better?

CMakeLists.txt
aminya: The benefit of using function arguments is that it allows building anything you want on top of it. For example, in case anyone wants to use global mutable states for managing their build, it is possible to add an option on top of the function arguemtns. 

Let's say for sanitizers:
```cmake
option(ENABLE_ADDRESS_SANITIZER_OPT "enable address sanitizer" OFF)

if(ENABLE_ADDRESS_SANITIZER_OPT)
	set(ENABLE_ADDRESS_SANITIZER ON)
else()
	set(ENABLE_ADDRESS_SANITIZER OFF)
endif()

cmakelib(
	# ...
	ENABLE_ADDRESS_SANITIZER 
)
```

ddalcino: I like the use of function arguments here, that's not what I'm objecting to. For me, much of the value of this project is that the boilerplate code is already written, and I don't have to go in and add all those cmake options myself. After this change, I will have very little reason to use this template at all; I might as well just use cmakelib instead and copy paste the boilerplate in the readme.

If this project is supposed to be a quick-start to having a C++ project, I think this change is moving directly against that goal. Maybe I'm misunderstanding something, maybe I'm being unnecessarily stubborn; in either case, I think we need input from @lefticus to resolve this

aminya: I don't understand why you think this is against the goal. Rust has 68000 packages that are all built using a single command `cargo build`. The reason is that everything is declarative by default. In contrast C++ only has 1000 packages on Conan.

Why do you think it is good to advertise using global mutable states and boilerplate code? I argue that declarative build system contributes to the reproducibility of the builds making it easier to write good packages. 

As I said, if anyone wants to do exotic things, they can do it. But I don't think we should advertise that. Also, Cmakelib is available for downloading. People can still download and modify that as they wish.

ClausKlein:
```cmake
option(ENABLE_ADDRESS_SANITIZER_OPT "enable address sanitizer" OFF)

if(ENABLE_ADDRESS_SANITIZER_OPT)
	set(ENABLE_ADDRESS_SANITIZER ON)
else()
	set(ENABLE_ADDRESS_SANITIZER OFF)
endif()

cmakelib(
	# ...
	ENABLE_ADDRESS_SANITIZER 
)
```
This can NOT work!

aminya: If you really want global options, we can add a file that can be optionally included to make those function arguments available as global options.

Here is a simple code that does the trick:
https://github.com/aminya/cmakelib/pull/9

ddalcino: > I don't understand why you think this is against the goal.

One of the stated goals of this project is to promote the use of sanitizers, coverage, static analyzers, etc. This is related to ["Use the tools available"](https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md). This change makes the tools unreachable from the command line, and requires code changes to turn the tools on and off.

Please keep in mind that a lot of people use this project (it has 1.4k stars at the moment), and depend on these features. They will not be pleased when you break them.

> I argue that declarative build system contributes to the reproducibility of the builds making it easier to write good packages.

This is a good point, and I agree with it. Using global options means that there will be variable name conflicts when one cpp_starter_project includes another cpp_starter_project. This is problem worth solving, and I don't have a ready solution for it.

I think it is really unproductive to continue to debate this without @lefticus' input. Even if I were to approve this PR, you will need his approval to merge, and it's hard for me to imagine that he would want to remove these features.

aminya: I just showed you a way to make things optionally global. Something like this

```cmake
# Make things global
include(${cmakelib_SOURCE_DIR}/GlobalOptions.cmake)

cmakelib(
  ${ENABLE_CACHE}
)
```

aminya: Fixed in https://github.com/aminya/cmakelib/pull/9

aminya:
@lefticus Any news? This is awaiting your review

build_examples.sh
aminya: The benefit of using Conan is that the code will be cross-platform. Qt itself is cross-platform, and using apt for installing a cross-platform library is not the best way. This means that the developer should write separate build scripts for every possible platform. This is the exact problem that Conan fixes. 

Also, Conan has a caching system, so there are no worries regarding the repeated builds.

Jason5480:
> There are certainly ways to make this work. One option would be to move them out of the `src` folder entirely, and into a parallel `examples` folder. Each example could have its own top-level CMakeLists.txt and conanfile.txt files; you could run `cmake -S ./examples/fltk ...` to build the fltk example. I don't really like this solution, but it is more declarative than what we have now, and it's better than leaving the examples inoperable. Perhaps you can come up with another solution that you like better?

Sorry for jumping out of the blue but @ddalcino the folder structure problem remind me an old CppCon talk https://www.youtube.com/watch?v=S4QSKLXdTtA&list=WL&t=1146s by Mateusz @mpusz. While a lot have change since then, the basic structure folder of src, test, and also you might add the example and fuzz_test folders under each sub-project is very convenient because then you have a CmakeLists.txt that include them all. In this talk Mateusz also uses several techniques and tricks like the include folder, the Alias libraries in Cmake and BUILD/INSTALL_INTERFACE that might be also useful here. I am aware that this repo is not for building a library and creating a conan package otherwise this would be called `cpp_starter_library` but there are ideas from the past that worth mentioning!

ddalcino:
> Sorry for jumping out of the blue but @ddalcino the folder structure problem remind me an old CppCon talk https://www.youtube.com/watch?v=S4QSKLXdTtA&list=WL&t=1146s by Mateusz @mpusz. While a lot have change since then, the basic structure folder of src, test, and also you might add the example and fuzz_test folders under each sub-project is very convenient because then you have a CmakeLists.txt that include them all. In this talk Mateusz also uses several techniques and tricks like the include folder, the Alias libraries in Cmake and BUILD/INSTALL_INTERFACE that might be also useful here. I am aware that this repo is not for building a library and creating a conan package otherwise this would be called `cpp_starter_library` but there are ideas from the past that worth mentioning!

@Jason5480 I remember that talk; it's a good one. Nested subprojects, reachable from the root level project, is a good approach in general, and in this case, it's the obvious one. Unfortunately, it will not work with this PR's `Conanfile.txt`: the SDL example would require that SDL be added to the Conanfile; the IMGUI example requires IMGUI, etc. We don't want to add SDL and IMGUI to the Conanfile if we're not building the examples.

The main branch of this project handles dependency management via `Conanfile.py`, which can add the required dependencies when they are needed. This PR removes the global options (`CPP_STARTER_USE_IMGUI`, `CPP_STARTER_USE_SDL`, etc) that are required to make this work. If we want to use Conan for dependency management, and keep the GUI examples in working condition, we need another solution.

We still need to be able to build everything in the project easily within one or two shell commands (see https://github.com/cpp-best-practices/cpp_starter_project/issues/147#issuecomment-904277057).

Jason5480:
> The main branch of this project handles dependency management via `Conanfile.py`, which can add the required dependencies when they are needed. This PR removes the global options (`CPP_STARTER_USE_IMGUI`, `CPP_STARTER_USE_SDL`, etc) that are required to make this work. If we want to use Conan for dependency management, and keep the GUI examples in working condition, we need another solution.
> 
> We still need to be able to build everything in the project easily within one or two shell commands (see [#147 (comment)](https://github.com/cpp-best-practices/cpp_starter_project/issues/147#issuecomment-904277057)).

Hmm I think I understand now. Examples produce different executables thus they are sub-projects that should have THEIR OWN conanfile.py. The installation of the dependencies should be done using conan_cmake_run() command in the CmakeLists.txt file of each example respectively. I believe that the root of the problem is that imgui, sdl and qt are not dependencies of the main project (because they are not used in the src/main.cpp) but they are dependencies of each example. So the cmake of each example should be responsible to install (through conanfile.py and conan_cmake_run) its own dependencies. Also I believe that fltk, gtkmm, imgui, nana, qt, sdl folders should be moved into a new cpp_starter_project/examples folder since they are not sources of the project.  The user should be able to chose if he want to build the examples but he should also be able to chose which gui framework to use for the project by modifying the main conanfile.py.

aminya:
I agree with moving the GUI projects to an examples folder if that helps remove the clutter. This current pull request has significantly simplified the starter_project. 

Many internal projects (which I work on) and were previously reluctant to use this template have switched to my declarative branch.

Now that the CppCon is over, @lefticus may be able to check out the pull request soon.

lefticus:
OK, I was able to give a brief look at this. I'm sorry for my absence, and unfortunately I'll have limited available again for the next 2 weeks.

I have a few thoughts:

1. I do love how composable it all looks
2. I don't love that with this setup that we cannot have the examples build, as others point out
3. I think the defaults are wrong

I know that I have not been paying close attention, so I would like to see reviews one way or the other from everyone who has been paying attention.

I think if we merge it, two things should happen:

1. We need to move the examples for GUI projects into a separate project
2. We need to make clang-tidy, cppcheck, other static analysis on by default, and force the user of the template to disable those
   * This is important since this is supposed to be a "Best Practices" template
   * It is also important for them to be enabled so that we can prove for ourselves that those options are actually working in our own project (is there a CMake coverage tool? :D )

ddalcino:
> 2. We need to make clang-tidy, cppcheck, other static analysis on by default, and force the user of the template to disable those

How about coverage? Is there a way to get coverage reports working if ENABLE_COVERAGE is turned off by default?

I agree with modifying `.clang-tidy` to reduce false positives. It’s much better to run it and not get all of its potential output than to leave it turned off. Our users can change their config files if they don’t like ours.

It would be nice if IWYU was turned on too, but IWYU can be very difficult to install. I’m not sure that it’s practical to have it turned on by default.

aminya:
> How about coverage? Is there a way to get coverage reports working if ENABLE_COVERAGE is turned off by default?
> 

Yes, we can do that. There is one potential improvement here. Our current coverage generation relies on a post-process step. I wish we could integrate gcovr and OpenCppCoverage into the actual CMake files.

> I agree with modifying `.clang-tidy` to reduce false positives. It’s much better to run it and not get all of its potential output than to leave it turned off. Our users can change their config files if they don’t like ours.

Agreed

> It would be nice if IWYU was turned on too, but IWYU can be very difficult to install. I’m not sure that it’s practical to have it turned on by default.

I'd rather not use IWYU. They have previously refused to make binaries available for unknown reasons, and building it from the source is very complicated (especially on Windows).

https://github.com/include-what-you-use/include-what-you-use/pull/799

lefticus:
> > We need to move the examples for GUI projects into a separate project
> 
> What is your suggestion for moving the GUI examples? Should it be another repository? If that's the case, I also think we should potentially add some more examples to that repository. For example, a lot of people use Eigen, MPI, OpenCV, Boost and other famous packages to build things. But we have not considered those at all.
> 

Yes, I imagine this project as the clean baseline one that doesn't require a bunch of stuff to be deleted, then we have a repository of examples that use this project and show how you would use it in real projects.

Does that make sense / sound reasonable? This is partially just for my own education, so I can understand how these pieces all related to each other.

aminya:
That sounds very good to me! Having a baseline encourages more people to use this template regardless of their usage (GUI, scientific programing, etc.) Then, we can have separate repositories for the examples.

Jason5480:
I agree with that proposal one template project for each use case!
Lets' capture the most common use cases of projects first. I would split that repo into:

- `cpp_simple_library_project`  where someone can develop a library using the best practices and create a conan package that can be reused. Variations like `cpp_sientific_library_project` `cpp_comm_library_project` that will use well known libraries that can be added for demonstration.
- `cpp_cli_project` where someone can create a command line application  like this one.
-  `cpp_qt_project` `cpp_imgui_project` where gui applications will be demonstrated using different frameworks (qt imgui etc.). Also they might consume  the libraries of the first bullet.

All those projects should re-use best practices techniques e.g. "CompilerWarnings.cmake" should be downloaded from an other repo by all aforementioned repositories just like conan.cmake does.

aminya:
I have enabled clang-tidy and cppcheck by default, but if they are not installed, a warning will be issued. Now, `WARNINGS_AS_ERRORS` switch is used to convert all warnings to errors (including ProjectOptions warnings).

aminya:
@lefticus @ddalcino I have addressed the things you mentioned. Fur sure, we can continue the development in other pull requests, but in my opinion, this PR is ready and I have been using it in production.

ddalcino:
> @lefticus @ddalcino I have addressed the things you mentioned. Fur sure, we can continue the development in other pull requests, but in my opinion, this PR is ready and I have been using it in production.

I agree that it makes sense to merge this before we get the examples working, because they should belong to their own separate repositories.

I can see that code coverage is working now. However, I'm not clear on why it works. Previously, we had the `ENABLE_COVERAGE` option controlling whether or not the `--coverage` flag is passed to the compiler. I see that you are using `ProjectOptions_ENABLE_COVERAGE` to gate the code that adds that flag, but it doesn't look like that option is set in the GH Action workflow. Is the `--coverage` flag used in that workflow?

There are a few issues that I can spot in [the last workflow run](https://github.com/cpp-best-practices/cpp_starter_project/pull/169/checks#step:6:16) that I am concerned about:
1. Warning about CMAKE_CXX_STANDARD:
    ```
    CMake Warning at build/_deps/projectoptions-src/src/StandardProjectSettings.cmake:45 (message):
      The default CMAKE_CXX_STANDARD used by external tools like clang-tidy and
      cppcheck is not set yet.  Using the latest supported C++ standard that is
      20
    ```
    We explicitly set the C++ standard to 17 in [the root level CMakeLists.txt at line 48](https://github.com/aminya/cpp_starter_project/blob/cmakelib/CMakeLists.txt#L48), but this warning is issued as a result of [the ProjectOptions call at line 25](https://github.com/aminya/cpp_starter_project/blob/cmakelib/CMakeLists.txt#L25). There's [a note in `ProjectOptions/src/StandardProjectSettings.cmake`](https://github.com/aminya/ProjectOptions/blob/main/src/StandardProjectSettings.cmake#L31-L34) that makes it sound like the user is supposed to set CMAKE_CXX_STANDARD before calling ProjectOptions. If at all possible, I would like `cpp_starter_project` to serve as an example of using ProjectOptions correctly without any warnings. I would not like our users to think "If the base project has these warnings, it's OK if my project does too. They must not mean anything important."
2. There are also warnings about `cppcheck` and `clang-tidy` not being present, which means that these tools will not run in CI. Since these tools are turned on by default, I would like to see them installed in the GH Actions workflow. The fact that they are not running in CI looks like a serious bug to me. I hope it's an easy fix.
3. After building spdlog: https://github.com/cpp-best-practices/cpp_starter_project/pull/169/checks#step:6:483
    ```
    ERROR: Not able to automatically detect '/usr/bin/c++' version
    ERROR: Unable to find a working compiler
    ```
    I genuinely do not understand what's happening here. I've never seen this before, and I imagine that it could cause some issues. Any ideas?
4. At the end: https://github.com/cpp-best-practices/cpp_starter_project/pull/169/checks#step:6:509
    ```
     Manually-specified variables were not used by the project:

        ENABLE_COVERAGE
    ```
    If this option no longer exists, shouldn't we remove it from the workflow?

There are several other warnings that I am not concerned about, like `ccache` is missing and `CMAKE_EXPORT_NO_PACKAGE_REGISTRY` specified but not used. At some point, it might be worth looking into these, but they are not related to this PR so there's no need to handle them now.

ddalcino:
@lefticus: Can we make some forward progress on moving ProjectOptions into the `cpp_best_practices` organization? The biggest reservation I have about approving this PR is that it moves most of the project into @aminya's personal repository. It's not that I don't trust @aminya; I feel that it is inappropriate. I intend to approve the PR, but I cannot do this until ProjectOptions becomes `cpp_best_practices/ProjectOptions`.

aminya:
I need to become a member of the whole organization (not just this repository), so I can transfer the project.

lefticus:
I just gave you both admin privs on a new project_options repo. That was the quick way to do this until I take a closer look at smarter organization options. Sorry for the delay. @ddalcino and @aminya

lefticus:
I went with project_options for consistency, but I realize now that transferring your repo first, then renaming it if that's the best thing to do makes more sense.

I'll delete the existing one, sorry about that. You should both be members now.

aminya:
Thanks for the invitation. I have transferred the project. 

I was also thinking about renaming to `project_options` for consistency with built-in CMake functions. I will do this after addressing the issues in [this comment](https://github.com/cpp-best-practices/cpp_starter_project/pull/169#issuecomment-986275094)

aminya:
> > @lefticus @ddalcino I have addressed the things you mentioned. Fur sure, we can continue the development in other pull requests, but in my opinion, this PR is ready and I have been using it in production.
> 
> I agree that it makes sense to merge this before we get the examples working, because they should belong to their own separate repositories.
> 
> I can see that code coverage is working now. However, I'm not clear on why it works. Previously, we had the `ENABLE_COVERAGE` option controlling whether or not the `--coverage` flag is passed to the compiler. I see that you are using `ProjectOptions_ENABLE_COVERAGE` to gate the code that adds that flag, but it doesn't look like that option is set in the GH Action workflow. Is the `--coverage` flag used in that workflow?
> 

I removed it in the recent commit.

> There are a few issues that I can spot in [the last workflow run](https://github.com/cpp-best-practices/cpp_starter_project/pull/169/checks#step:6:16) that I am concerned about:
> 
> 1. Warning about CMAKE_CXX_STANDARD:
>    ```
>    CMake Warning at build/_deps/projectoptions-src/src/StandardProjectSettings.cmake:45 (message):
>      The default CMAKE_CXX_STANDARD used by external tools like clang-tidy and
>      cppcheck is not set yet.  Using the latest supported C++ standard that is
>      20
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    We explicitly set the C++ standard to 17 in [the root level CMakeLists.txt at line 48](https://github.com/aminya/cpp_starter_project/blob/cmakelib/CMakeLists.txt?rgh-link-date=2021-12-05T18%3A08%3A22Z#L48), but this warning is issued as a result of [the ProjectOptions call at line 25](https://github.com/aminya/cpp_starter_project/blob/cmakelib/CMakeLists.txt?rgh-link-date=2021-12-05T18%3A08%3A22Z#L25). There's [a note in `ProjectOptions/src/StandardProjectSettings.cmake`](https://github.com/aminya/ProjectOptions/blob/main/src/StandardProjectSettings.cmake?rgh-link-date=2021-12-05T18%3A08%3A22Z#L31-L34) that makes it sound like the user is supposed to set CMAKE_CXX_STANDARD before calling ProjectOptions. If at all possible, I would like `cpp_starter_project` to serve as an example of using ProjectOptions correctly without any warnings. I would not like our users to think "If the base project has these warnings, it's OK if my project does too. They must not mean anything important."


This is now changed to STATUS instead of WARNING. I agree that this does not need to be a warning. 


> 2. There are also warnings about `cppcheck` and `clang-tidy` not being present, which means that these tools will not run in CI. Since these tools are turned on by default, I would like to see them installed in the GH Actions workflow. The fact that they are not running in CI looks like a serious bug to me. I hope it will be easy to install them.

This is intended and it is not a bug. Now, to make these errors instead of warnings, `WARNINGS_AS_ERRORS` should be enabled. The compiler, clang-tidy, and cppcheck warnings are now treated equally. This also encourages setting these options on by default regardless since the end-users of the library do not need to have these tools installed to build the library for production. 


> 3. After building spdlog: [#169 (checks)](https://github.com/cpp-best-practices/cpp_starter_project/pull/169/checks#step:6:483)
>    ```
>    ERROR: Not able to automatically detect '/usr/bin/c++' version
>    ERROR: Unable to find a working compiler
>    ``` 
>      
>    I genuinely do not understand what's happening here. I've never seen this before, and I imagine that it could cause some issues. Any ideas?

These seem to be something internal to the spdlog package. I haven't seen anything similar in the projects I have used.

CMakeLists.txt
```diff
@@ -1,86 +1,66 @@
...
+option(ENABLE_TESTING "Enable the tests" ON)
 if(ENABLE_TESTING)
   enable_testing()
-  message("Building Tests. Be sure to check out test/constexpr_tests for constexpr testing")
+  message("Building Tests. Be sure to check out test/constexpr_tests for constexpr
+testing")
   add_subdirectory(test)
 endif()
 
+option(ENABLE_FUZZING "Enable the fuzz tests" OFF)
```
ddalcino, line 62: This PR removes all of the CMake options from the project, except for these two: ENABLE_TESTING and ENABLE_FUZZING. I am not objecting to this; I am just curious to know why these options were not removed. Given the previous decisions made in this PR, I would have expected these options to disappear, and for the `enable_testing()` and `add_subdirectory()` calls to run by default.

Theoretically, given 2 projects derived from `cpp_starter_project`, where one project is dependent on the other project, will there be a name collision between the two `ENABLE_TESTING` options? I have not tried to do this myself, but I think it would be worthwhile to know the answer. I thought this was representative of the problem we were trying to avoid by removing all of the other options.

build_examples.sh
```diff
@@ -1,133 +1,42 @@
 #!/usr/bin/env bash
 
-# This script is intended to be run from an Ubuntu 18.04 Docker image.
-# Since it uses `apt-get` to install dependencies required by the examples,
-# it may also run on other versions of Ubuntu, or other Debian-derived distros.
-# It will definitely not run without `apt-get` installed!
+# This script is intended to be run from an Ubuntu Docker image (or other Debian-derived distros).
+# It uses `apt-get` to install dependencies required by the examples,
```
ddalcino, line 4: Since the GUI examples are moving into their own repos, maybe this file should just be removed entirely?

src/CMakeLists.txt
```diff
@@ -1,45 +1,22 @@
...
-  add_subdirectory(imgui)
-endif()
+# add_subdirectory(imgui)
 
 # Nana
-if(CPP_STARTER_USE_NANA)
-  message("Using Nana")
-  add_subdirectory(nana)
-endif()
+# add_subdirectory(nana)
```
ddalcino, line 19: Since the GUI examples are moving into their own repos, maybe all the comments from lines 0-19 should be removed? Should we delete the `src/nana`, `imgui`, `gtkmm` etc. directories as well?

ddalcino (approved):
This looks good enough now that I don't think I should stand in the way.

My approval assumes that the following issues will be resolved separately from this PR:

1. Install ccache, cppcheck, and clang-tidy in CI. I think setup-cpp is a good candidate for this task, but I am not opposed to installing these tools with explicit shell commands in the workflow.
2. Add separate repositories to the `cpp_best_practices` org for the existing GUI examples.

aminya:
Renaming the package is blocked:
https://github.com/cpp-best-practices/ProjectOptions/pull/37

build_examples.sh
aminya: Yes. I think it is good to address these separately.

src/CMakeLists.txt
aminya: Yes, eventually will do.

CMakeLists.txt
aminya: The tests are a bit different than the GUI examples. Because tests are always written for all the projects, and constantly disabling/enabling them is not desirables. That's why this is a case where using an option is not a bad idea.
If CMake had a standardized variable for enabling and building the tests, then we did not have this issue here. In Rust, people will use `cargo test` to build and run the tests, and everything happens automatically. Because CMake lacks such command, we can treat `cmake -S ./ -B ./build -DENABLE_TESTING` like the build part of `cargo test`

aminya:
The package is renamed now, and based on the tests, it seems it works as it should. This should be ready to go.

CMakeLists.txt
```diff
@@ -1,86 +1,66 @@
...
+# uncomment to enable vcpkg:
+# # Setup vcpkg - should be called before defining project()
+# run_vcpkg()
+
+# Set the project name and language
+project(myproject LANGUAGES CXX)
+
+# Initialize project_options
+# uncomment the options to enable them:
+project_options(
```
ddalcino, line 25: Apologies in advance if this critique sounds nit-picky, because it is definitely nit-picky.

I think we have a name collision here. As I understand it, we have two variables named `project_options`:
* the macro called `project_options`, that we imported from the `project_options` library
* the 'library interface' called `project_options`, that was created when we called the `project_options` macro

To my understanding, the library interface `project_options` object will replace the macro `project_options`, and every time you reference `project_options` after this point, you will get the library interface and not the macro. This should work just fine, and it will not cause any issue unless you decide you want to use the `project_options` macro again.

I think that this could be very difficult for a first-time user of the project to understand. To understand this code, you are forced to dig into the 'project_options' source code. I think it would be easier to understand if the names of the macro and the library interface were different, to make it clear that they are not the same thing.

I think the `project_options` macro could be easier to understand if it allowed the user to set the name of the library interface, like this:

```cmake
# create `my_project_options_interface_target`
project_options(                         # call the macro
  my_project_options_interface_target    # name of the library interface to create
  ENABLE_CACHE                           # option to turn on
)

# use `my_project_options_interface_target`
target_compile_features(my_project_options_interface_target INTERFACE cxx_std_17)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE my_project_options_interface_target)
```

CMakeLists.txt
```diff
@@ -1,86 +1,66 @@
...
+      # ENABLE_UNITY
+      # ENABLE_SANITIZER_ADDRESS
+      # ENABLE_SANITIZER_LEAK
+      # ENABLE_SANITIZER_UNDEFINED_BEHAVIOR
+      # ENABLE_SANITIZER_THREAD
+      # ENABLE_SANITIZER_MEMORY
+      # CONAN_OPTIONS
+)
+# project_options is defined inside cmakelib
 target_compile_features(project_options INTERFACE cxx_std_17)
```
ddalcino, line 48: Could it be possible to pass `cxx_std_17` as a parameter to the `project_options` macro? That would make it so that there's no need to modify the `project_options` library interface after it has been created. It could also help with setting a default value for `CMAKE_CXX_STANDARD` in [`project_options/src/StandardProjectSettings.cmake`](https://github.com/cpp-best-practices/project_options/blob/d0db61776eb02128b8c65b16e509b6ffe82630ee/src/StandardProjectSettings.cmake#L49).

CMakeLists.txt
aminya: Thanks, I thought about it. But I am not sure if we want allow calling project_options twice the same way Cmake allows only one project.

CMakeLists.txt
ddalcino: > The tests are a bit different than the GUI examples. Because tests are always written for all the projects, and constantly disabling/enabling them is not desirables. That's why this is a case where using an option is the better solution.

This sounds like an argument for restoring several of the options that were removed. Constantly enabling any of the sanitizer options cannot possibly be desirable.

CMakeLists.txt
ddalcino: Maybe we could update the comment to explain what's going on, so that new users don't get the idea that this call to the `project_options()` macro is a constructor?

Also, the comment below this still refers to `cmakelib`.

conanfile.txt
```diff
@@ -0,0 +1,16 @@
...
+[requires]
+catch2/2.13.7
+docopt.cpp/0.6.2
+fmt/8.0.1
+spdlog/1.9.2
+# imgui-sfml/2.1@bincrafters/stable
+# sdl2/2.0.10@bincrafters/stable
+
+[generators]
+cmake_find_package
```
Jason5480, line 12: Are all those generators necessary? I would prefer the conan to be less intrusive to the CMakeList.txt files. There is the option to generate a toolchain file as described [here]( https://docs.conan.io/en/1.32/creating_packages/toolchains/cmake.html) and then pass it (if you want) to the CMake command. I am not familiar to the vcpkg but as I can see [here](https://vcpkg.readthedocs.io/en/latest/examples/installing-and-using-packages/) it also produce a toolchain .cmake file that can be passed as an `CMAKE_TOOLCHAIN_FILE` argument. This way the user can chose between conan and vcpkg at will, without the need for a package specific logic inside the CMakeLists.txt

Jason5480: Also I do not see any benefits of converting conanfile.py to conanfile.txt since all developer machines have python installed already and the capabilities of the .txt file are way less powerful than the .py

ddalcino: See https://github.com/cpp-best-practices/cpp_starter_project/pull/169#discussion_r715242962 and following comments. The .txt is declarative and does not require a developer to understand Python to debug it.

CMakeLists.txt
aminya: Fixed in https://github.com/cpp-best-practices/cpp_starter_project/pull/169/commits/4221006d809b899174588a7581e8ada2f0a53c4e

CMakeLists.txt
aminya: Setting a default `CMAKE_CXX_STANDARD` is a good practice. If not set, `project_options` simply set it to the latest version. This is consistent with the behaviour of other programming languages which default to the latest standard instead of some old compatibility standard.

conanfile.txt
aminya: These generators can be removed if are not needed. There is no noticeable overhead in supporting all of them, so I think there is no harm in specifying all of them. 

vcpkg is already supported using `run_vcpkg` function.

CMakeLists.txt
aminya: The global options are still available if you need them. Include `GlobalOptions.cmake`

aminya:
What else is holding this back?

lefticus:
Looks like it's probably time for me to catch back up on this...

lefticus (approved):
Looks clean and good, thank you.

aminya:
> Looks clean and good, thank you.

You are welcome!

Issue #129, opened 2021-03-12
Add CI tests for GUI examples


ddalcino:
The current CI builds do not check to see if the GUI examples build successfully. It would be really helpful for the CI to tell us whether or not a change would break the examples.

While writing #121 I noticed that both the Nana and Qt examples produce warnings, which lead to build failures, unless you define `WARNINGS_AS_ERRORS=FALSE`. As a result, I had to add some code to #121 that really doesn't belong there, just to prove that my branch works. CI tests would have revealed these problems a long time ago:
* The Qt example fails because of a stray semicolon at `src/qt/HelloQt.cpp` line 38.
* The Nana example fails because the Nana header files have some shadowed variable names, which triggers `-Wshadow`.

Building the examples should be as frictionless as possible, and CI builds would really help with that. What better way to destroy user confidence is there than to provide examples that don't work?

If anyone is interested in adding these CI tests, you may find it useful to read the `build_examples.sh` and `Dockerfile` files in #121, or the `Dockerfile` in #99. Building the examples is not as straightforward as I would like it to be.

**Edit**: Since #134 and #121 were merged, the compilation failures described here should be resolved.

lefticus:
@ddalcino I'm closing this since we have commented out (and then deleted) most of the un-tested GUI demos.

I'm slowly re-adding what I can.

Notes:

 * conan support for Qt is spotty
 * SFML cannot build with modern compilers because it still uses `auto_ptr` (removed in C++17)
 * FLTK - possible to get working again
 * nana - cannot compile in C++20 mode, so I might end up creating another project just to demo it?

PR #170, opened 2021-09-23
modernize cmake a bit

ClausKlein:
add conan, CPM, and catch commands to .cmake-format.yaml file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,10 +78,6 @@ if(ENABLE_FUZZING)
   add_subdirectory(fuzz_test)
 endif()
 
-add_subdirectory(src)
-
 option(ENABLE_UNITY "Enable Unity builds of projects" ON)
-if(ENABLE_UNITY)
-  # Add for any project you want to apply unity builds for
-  set_target_properties(intro PROPERTIES UNITY_BUILD ON)
-endif()
+
+add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,6 @@ if(ENABLE_FUZZING)
   add_subdirectory(fuzz_test)
 endif()
 
-option(ENABLE_UNITY "Enable Unity builds of projects" ON)
+option(ENABLE_UNITY "Enable Unity builds of projects" OFF)
 
 add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
-cmake_minimum_required(VERSION 3.15...3.21)
+cmake_minimum_required(VERSION 3.16...3.21)
 
-# Set the project name to your project name, my project isn't very descriptive
+# TODO: Set the project name to your project name, and version; my project isn't very descriptive
 project(myproject VERSION 0.1.0 LANGUAGES CXX)
 
 include(cmake/StandardProjectSettings.cmake)
@@ -10,6 +10,11 @@ include(cmake/PreventInSourceBuilds.cmake)
 add_library(project_options INTERFACE)
 target_compile_features(project_options INTERFACE cxx_std_17)
 
+#
+# FIXME: TBC
+#  IMHO: should be moved to cmake/StandardProjectSettings.cmake CK
+#
+# XXX configure_compiler(project_options)
 if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
   option(ENABLE_BUILD_WITH_TIME_TRACE "Enable -ftime-trace to generate time tracing .json files on clang" OFF)
   if(ENABLE_BUILD_WITH_TIME_TRACE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,8 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.15...3.21)
 
 # Set the project name to your project name, my project isn't very descriptive
-project(myproject CXX)
+project(myproject VERSION 0.1.0 LANGUAGES CXX)
+
 include(cmake/StandardProjectSettings.cmake)
 include(cmake/PreventInSourceBuilds.cmake)
 
@@ -79,7 +80,7 @@ endif()
 
 add_subdirectory(src)
 
-option(ENABLE_UNITY "Enable Unity builds of projects" OFF)
+option(ENABLE_UNITY "Enable Unity builds of projects" ON)
 if(ENABLE_UNITY)
   # Add for any project you want to apply unity builds for
   set_target_properties(intro PROPERTIES UNITY_BUILD ON)
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -40,7 +40,7 @@ macro(run_conan)
     # the external "conanfile.py" provided with the project
     # Alternatively a conanfile.txt could be used
     conan_cmake_install(
-      ERROR_QUIET
+      OUTPUT_QUIET
       PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR}
       BUILD missing
       # Pass compile-time configured options into conan
--- a/cmake/Conan.cmake
+++ b/cmake/Conan.cmake
@@ -1,15 +1,11 @@
 macro(run_conan)
   # Download automatically, you can also just copy the conan.cmake file
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/conan.cmake")
-    message(
-      STATUS
-        "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
+    message(STATUS "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
     file(
-      DOWNLOAD
-      "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
+      DOWNLOAD "https://raw.githubusercontent.com/conan-io/cmake-conan/v0.16.1/conan.cmake"
       "${CMAKE_BINARY_DIR}/conan.cmake"
-      EXPECTED_HASH
-        SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
+      EXPECTED_HASH SHA256=396e16d0f5eabdc6a14afddbcfff62a54a7ee75c6da23f32f7a31bc85db23484
       TLS_VERIFY ON)
   endif()
 
@@ -17,14 +13,13 @@ macro(run_conan)
   list(APPEND CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR})
   list(APPEND CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR})
 
+  # XXX see i.e.: cmake-build-cpp_starter_project-x86_64-Debug/install/conan.cmake
   include(${CMAKE_BINARY_DIR}/conan.cmake)
 
   # Add (or remove) remotes as needed
-  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
+  ### conan_add_remote(NAME conan-center URL https://conan.bintray.com)
   conan_add_remote(NAME cci URL https://center.conan.io INDEX 0)
-  conan_add_remote(
-    NAME bincrafters URL
-    https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
+  conan_add_remote(NAME bincrafters URL https://bincrafters.jfrog.io/artifactory/api/conan/public-conan)
 
   # For multi configuration generators, like VS and XCode
   if(NOT CMAKE_CONFIGURATION_TYPES)
@@ -43,18 +38,14 @@ macro(run_conan)
 
     # PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR} is used to tell conan to process
     # the external "conanfile.py" provided with the project
-    # Alternatively a conanfile.txt could be used 
+    # Alternatively a conanfile.txt could be used
     conan_cmake_install(
-      PATH_OR_REFERENCE
-      ${CMAKE_SOURCE_DIR}
-      BUILD
-      missing
+      ERROR_QUIET
+      PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR}
+      BUILD missing
       # Pass compile-time configured options into conan
-      OPTIONS
-      cpp_starter_use_imgui=${CPP_STARTER_USE_IMGUI}
-      cpp_starter_use_sdl=${CPP_STARTER_USE_SDL}
-      SETTINGS
-      ${settings})
+      OPTIONS cpp_starter_use_imgui=${CPP_STARTER_USE_IMGUI} cpp_starter_use_sdl=${CPP_STARTER_USE_SDL}
+      SETTINGS ${settings})
   endforeach()
 
 endmacro()
--- a/cmake/StandardProjectSettings.cmake
+++ b/cmake/StandardProjectSettings.cmake
@@ -32,6 +32,8 @@ if(ENABLE_IPO)
     message(SEND_ERROR "IPO is not supported: ${output}")
   endif()
 endif()
+
+# TODO: do not use global add_compiler_options()! CK
 if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
   add_compile_options(-fcolor-diagnostics)
 elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2,7 +2,7 @@
 // The full include-list reported from include-what-you-use
 //
 #include <docopt/docopt.h>  // for docopt
-#include <docopt_value.h>   // for operator<<, value
+#include <docopt/docopt_value.h>   // for operator<<, value
 #include <fmt/format.h>     // for make_format_args, print, vformat_to
 #include <spdlog/spdlog.h>  // for info
 #include <exception>        // for exception
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -14,7 +14,7 @@ static constexpr auto USAGE =
           naval_fate mine (set|remove) <x> <y> [--moored | --drifting]
           naval_fate (-h | --help)
           naval_fate --version
- Options:
+    Options:
           -h --help     Show this screen.
           --version     Show version.
           --speed=<kn>  Speed in knots [default: 10].
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,8 +1,16 @@
-#include <functional>
-#include <iostream>
-
-#include <spdlog/spdlog.h>
-#include <docopt/docopt.h>
+//
+// The full include-list reported from include-what-you-use
+//
+#include <docopt/docopt.h>  // for docopt
+#include <docopt_value.h>   // for operator<<, value
+#include <fmt/format.h>     // for make_format_args, print, vformat_to
+#include <spdlog/spdlog.h>  // for info
+#include <exception>        // for exception
+#include <iostream>         // for string, char_traits, operator<<, basic_os...
+#include <iterator>         // for next
+#include <map>              // for map, __map_iterator, operator!=
+#include <type_traits>      // for enable_if<>::type
+#include <utility>          // for pair
 
 static constexpr auto USAGE =
   R"(Naval Fate.
@@ -22,7 +30,7 @@ static constexpr auto USAGE =
           --drifting    Drifting mine.
 )";
 
-int main()
+int main([[maybe_unused]] int argc, [[maybe_unused]] char *argv[])
 {
   std::map<std::string, docopt::value> args = docopt::docopt(USAGE,
     { std::next(argv), std::next(argv, argc) },
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -22,7 +22,7 @@ static constexpr auto USAGE =
           --drifting    Drifting mine.
 )";
 
-int main(int argc, const char **argv)
+int main()
 {
   std::map<std::string, docopt::value> args = docopt::docopt(USAGE,
     { std::next(argv), std::next(argv, argc) },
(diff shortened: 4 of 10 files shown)


aminya (changes requested):
This should await #169.

CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.15...3.21)
```
ClausKlein, line 1: for[UNITY_BUILD]( https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html) it should be `3.19...3.21`

src/sdl/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
+find_package(SDL2 CONFIG REQUIRED)
+
 add_executable(test_sdl main.cpp)
+target_link_libraries(test_sdl PRIVATE SDL2::SDL2)
+target_link_libraries(test_sdl PRIVATE project_warnings project_options)
+set_target_properties(test_sdl PROPERTIES UNITY_BUILD ${ENABLE_UNITY})
+
+# NOTE: this quickfix was needed at leased with brew on OSX! CK
+# target_include_directories(test_sdl PRIVATE ${SDL2_INCLUDE_DIRS}/..)
```
ClausKlein, line 9: **NOTE**: before my change, it does not compile on OSX

src/sdl/main.cpp
```diff
@@ -40,7 +46,7 @@ static void check_video_driver(const char *name)
 }
 
 
-int main(int argc, char *argv[]) {
+int main(int /* argc */, char ** /* argv */) {
```
ddalcino, line 49: This project requires C++17 as a minimum, which means you have access to `[[maybe_unused]]` here. Since this project emphasizes best-practices, I think it makes more sense here to use the attribute rather than comment out the variable names.
```suggestion
int main([[maybe_unused]] int argc, [[maybe_unused]] char ** argv) {
```

test/CMakeLists.txt
```diff
@@ -14,33 +14,23 @@ target_link_libraries(tests PRIVATE project_warnings project_options catch_main)
...
-  .
-  OUTPUT_PREFIX
-  "unittests."
-  OUTPUT_SUFFIX
-  .xml)
+  TEST_PREFIX "unittests."
+  REPORTER xml
+  OUTPUT_DIR .
+  OUTPUT_PREFIX "unittests."
+  OUTPUT_SUFFIX .xml)
```
ddalcino, line 21: This is a huge improvement in terms of readability, IMHO. Thank you for finding the `.cmake-format.yaml` parameters that permit this!

src/sdl/CMakeLists.txt
ddalcino: Good catch. This underscores the need to do something about #129

cmake/Conan.cmake
```diff
@@ -43,18 +38,14 @@ macro(run_conan)
...
     # PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR} is used to tell conan to process
     # the external "conanfile.py" provided with the project
-    # Alternatively a conanfile.txt could be used 
+    # Alternatively a conanfile.txt could be used
     conan_cmake_install(
-      PATH_OR_REFERENCE
-      ${CMAKE_SOURCE_DIR}
-      BUILD
-      missing
+      ERROR_QUIET
```
ddalcino, line 43: I am unclear on the reason you're adding `ERROR_QUIET` here. As I understand the [docs](https://cmake.org/cmake/help/latest/command/execute_process.html?highlight=error_quiet), `ERROR_QUIET` silences Conan's output to stderr. I think this could make it very difficult to figure out what went wrong. Can you explain your reasoning here?

CMakeLists.txt
ddalcino: I am unclear on the purpose of the [`<policy_max>`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) parameter here. 

Broadly speaking, I envision two kinds of people using this project:
1. Users like yourself who know CMake well enough to know the exact range of CMake versions that will work for their project. These users will know about the `<policy_max>` parameter, and will set it to the appropriate value for themselves.
2. Users who don't know so much about CMake yet, and will be very frustrated when they try to compile this project for the first time, because their package manager has installed CMake 3.22.

I do not understand how setting `<policy_max>` in the starter project is of benefit to either user. Can you please provide some more context here?

cmake/Conan.cmake
ClausKlein: When I started testen, `canon` and `cmake` reported errors I could not understand. 
To analyse it, I added this option.

src/sdl/main.cpp
ClausKlein: yes, and no.

In this case, the parameters are never used.

I would prefer to use
```c++
int main() {
```

CMakeLists.txt
ClausKlein: The max version is the highest policy used be `cmake` if a newer version of `cmake` is used.

I have tested with `3.21`, so I can guaranty this `cmake build system` work with it.
but you need at leased `3.16` to use the `UNITY_BUILD` property!

cmake/Conan.cmake
ClausKlein: this was the output:

```bash
----Running------
> cd '/Users/clausklein/.conan/data/docopt.cpp/0.6.2/_/_/build/2f6fd32d99347efdec5280d9a5c607e18b4e39e1/build_subfolder' &&
> cmake -G "Ninja" -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_OSX_ARCHITECTURES="x86_64"
> -DCMAKE_OSX_SYSROOT="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk"
> -DCONAN_IN_LOCAL_CACHE="ON" -DCONAN_COMPILER="clang" -DCONAN_COMPILER_VERSION="12" -DCONAN_CXX_FLAGS="-m64"
> -DCONAN_SHARED_LINKER_FLAGS="-m64" -DCONAN_C_FLAGS="-m64" -DCONAN_LIBCXX="libc++" -DBUILD_SHARED_LIBS="ON"
> -DCMAKE_INSTALL_PREFIX="/Users/clausklein/.conan/data/docopt.cpp/0.6.2/_/_/package/2f6fd32d99347efdec5280d9a5c607e18b4e39e1"
> -DCMAKE_INSTALL_BINDIR="bin" -DCMAKE_INSTALL_SBINDIR="bin" -DCMAKE_INSTALL_LIBEXECDIR="bin" -DCMAKE_INSTALL_LIBDIR="lib"
> -DCMAKE_INSTALL_INCLUDEDIR="include" -DCMAKE_INSTALL_OLDINCLUDEDIR="include" -DCMAKE_INSTALL_DATAROOTDIR="share"
> -DCMAKE_MODULE_PATH="/Users/clausklein/.conan/data/docopt.cpp/0.6.2/_/_/build/2f6fd32d99347efdec5280d9a5c607e18b4e39e1"
> -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY="ON" -DCONAN_EXPORTED="1" -DUSE_BOOST_REGEX="False" -Wno-dev
> '/Users/clausklein/.conan/data/docopt.cpp/0.6.2/_/_/build/2f6fd32d99347efdec5280d9a5c607e18b4e39e1'
-----------------
Logging command output to file '/Users/clausklein/.conan/data/docopt.cpp/0.6.2/_/_/build/2f6fd32d99347efdec5280d9a5c607e18b4e39e1/conan_run.log'
CMake Error at /usr/local/Cellar/cmake/3.21.3/share/cmake/Modules/CMakeDetermineCCompiler.cmake:49 (message):
  Could not find compiler set in environment variable CC:

  /usr/local/bin/gcc-10.
  CMakeLists.txt:2 (project)


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!

```

ddalcino: If `conan_install` failed, should not CMake fail as well? In this case `ERROR_QUIET` allows CMake to continue when it should probably fail.

If you are able to get this project running at all, and the Conan process is failing, you must have installed the dependencies independently of Conan. If this is the case, you can turn off Conan with the `ENABLE_CONAN` CMake option.

The `Could not find compiler set in environment variable CC` looks like a configuration problem; it is not the expected result, and will not result in a properly built binary.

CMakeLists.txt
ddalcino:
```suggestion
cmake_minimum_required(VERSION 3.16)
```
Please remove the `<policy_max>` parameter. Unless you have a good reason to use it, it will break our users' code when they upgrade CMake. You don't have to guarantee that your change will work with future versions of CMake.

src/sdl/main.cpp
ddalcino: > I would prefer to use
> 
> ```c++
> int main() {
> ```
I agree, that looks much better.

cmake/Conan.cmake
aminya: To remove noise add `OUTPUT_QUIET`. `ERROR_QUIET` should not be used as it hides the errors

CMakeLists.txt
ClausKlein: > ...  it will break our users' code when they upgrade `CMake`. You don't have to guarantee that your change will work with future versions of CMake.

That is simply not true!

I can guaranty it for `CMake 3.21.3` And `CMake ` does it with this concept: https://cmake.org/cmake/help/v3.21/manual/cmake-policies.7.html?highlight=policy

src/sdl/main.cpp
ClausKlein: ok

cmake/Conan.cmake
ClausKlein: ok

ClausKlein:
What is the problem with conan on my build host?
```
-- Conan executing: /usr/local/bin/conan install /Users/clausklein/Workspace/cpp/cpp_starter_project --build missing --options cpp_starter_use_imgui=OFF --options cpp_starter_use_sdl=ON --settings build_type=Release --settings compiler=gcc --settings compiler.version=11 --settings compiler.libcxx=libstdc++11
sdl2/2.0.10@bincrafters/stable: WARN: Remote https://conan.bintray.com is deprecated and will be shut down soon.
sdl2/2.0.10@bincrafters/stable: WARN: Please use the new 'conancenter' default remote.
sdl2/2.0.10@bincrafters/stable: WARN: Add it to your remotes with: conan remote add -i 0 conancenter https://center.conan.io
ERROR: Failed requirement 'sdl2/2.0.10@bincrafters/stable' from 'conanfile.py (CppStarterProject/0.1)'
ERROR: Permission denied for user: 'clausklein'. [Remote: pocoproject]
CMake Error at cmake-build-cpp_starter_project-x86_64-Release/install/conan.cmake:631 (message):
  Conan install failed='1'
  cmake/Conan.cmake:42 (conan_cmake_install)
  CMakeLists.txt:72 (run_conan)


-- Configuring incomplete, errors occurred!
See also "/Users/clausklein/Workspace/cpp/cpp_starter_project/cmake-build-cpp_starter_project-x86_64-Release/install/CMakeFiles/CMakeOutput.log".
bash-3.2$ 
```

CMakeLists.txt
```diff
@@ -1,14 +1,20 @@
...
 
 # Link this 'library' to set the c++ standard / compile-time options requested
 add_library(project_options INTERFACE)
 target_compile_features(project_options INTERFACE cxx_std_17)
 
+#
+# FIXME: TBC
+#  IMHO: should be moved to cmake/StandardProjectSettings.cmake CK
+#
+# XXX configure_compiler(project_options)
```
ClausKlein, line 17: what do you mean about this?

cmake/StandardProjectSettings.cmake
```diff
@@ -32,6 +32,8 @@ if(ENABLE_IPO)
     message(SEND_ERROR "IPO is not supported: ${output}")
   endif()
 endif()
+
+# TODO: do not use global add_compiler_options()! CK
```
ClausKlein, line 36: please note

src/sdl/CMakeLists.txt
ClausKlein: should I remove my comments?

src/sdl/main.cpp
ClausKlein: this break the `travis` build with `MVC`?

ClausKlein:
Without `conan` the project does not build, indeed I have install `docopt` project:

```bash
cmake --build ./cmake-build-docopt.cpp-x86_64-Release/install --target install
[0/1] Install the project...
-- Install configuration: "Release"
-- Installing: /usr/local/lib/libdocopt.0.6.2.dylib
-- Up-to-date: /usr/local/lib/libdocopt.0.dylib
-- Up-to-date: /usr/local/lib/libdocopt.dylib
-- Installing: /usr/local/lib/libdocopt.a
-- Up-to-date: /usr/local/include/docopt/docopt.h
-- Up-to-date: /usr/local/include/docopt/docopt_private.h
-- Up-to-date: /usr/local/include/docopt/docopt_util.h
-- Up-to-date: /usr/local/include/docopt/docopt_value.h
-- Up-to-date: /usr/local/lib/cmake/docopt/docopt-config.cmake
-- Installing: /usr/local/lib/cmake/docopt/docopt-config-version.cmake
-- Installing: /usr/local/lib/cmake/docopt/docopt-targets.cmake
-- Installing: /usr/local/lib/cmake/docopt/docopt-targets-release.cmake
-- Installing: /usr/local/lib/pkgconfig/docopt.pc
rm -rf ./cmake-build-docopt.cpp-x86_64-Release/install
bash-3.2$ 

# ...
cmake -S . -B ./cmake-build-cpp_starter_project-x86_64-Release/install -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/usr/local -DENABLE_INCLUDE_WHAT_YOU_USE=1 -DCPP_STARTER_USE_SDL=ON -DENABLE_CONAN=0 -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_CXX_STANDARD=17
-- ccache found and enabled
Building Tests. Be sure to check out test/constexpr_tests for constexpr testing
Using SDL2
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Configuring done
CMake Error at src/CMakeLists.txt:49 (add_executable):
  Target "intro" links to target "docopt::docopt" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
bash-3.2$ 
```

src/main.cpp
```diff
@@ -1,8 +1,16 @@
-#include <functional>
-#include <iostream>
-
-#include <spdlog/spdlog.h>
-#include <docopt/docopt.h>
+//
+// The full include-list reported from include-what-you-use
+//
+#include <docopt/docopt.h>  // for docopt
+#include <docopt_value.h>   // for operator<<, value
```
ClausKlein, line 5:
```cmd
cmake --build ./build --config "Release"
143Microsoft (R) Build Engine version 16.11.0+0538acc04 for .NET Framework
144Copyright (C) Microsoft Corporation. All rights reserved.
145
146  Checking Build System
147  Building Custom Rule C:/projects/source/test/CMakeLists.txt
148  catch_main.cpp
149  catch_main.vcxproj -> C:\projects\source\build\test\Release\catch_main.lib
150  Building Custom Rule C:/projects/source/test/CMakeLists.txt
151  constexpr_tests.cpp
152  constexpr_tests.vcxproj -> C:\projects\source\build\test\Release\constexpr_tests.exe
153  Building Custom Rule C:/projects/source/src/CMakeLists.txt
154  main.cpp
155C:\projects\source\src\main.cpp(5,10): fatal error C1083: Cannot open include file: 'docopt_value.h': No such file or directory [C:\projects\source\build\src\intro.vcxproj]
156  Building Custom Rule C:/projects/source/test/CMakeLists.txt
157  constexpr_tests.cpp
158  relaxed_constexpr_tests.vcxproj -> C:\projects\source\build\test\Release\relaxed_constexpr_tests.exe
159  Building Custom Rule C:/projects/source/test/CMakeLists.txt
160  tests.cpp
161  tests.vcxproj -> C:\projects\source\build\test\Release\tests.exe
162
```

ClausKlein (commented):
it is a nightmare with **MVC**!

src/sdl/main.cpp
ClausKlein: fixed

src/main.cpp
ClausKlein: fixed

CMakeLists.txt
ddalcino: > > ...  it will break our users' code when they upgrade `CMake`. You don't have to guarantee that your change will work with future versions of CMake.
> 
> That is simply not true!

You are correct; this assessment was premature. `<policy_max>` is intended to ensure forwards compatibility with future versions of CMake, not to break it. This code will ensure that any user who configures the project with CMake 3.22+ will have the same experience as someone using CMake 3.21.

The part that I still don't understand is why you have selected 3.21 as the `<policy max>`. If it's just because 3.21 is the newest CMake currently available, then I think this could be the wrong decision. Will we need to bump this version number every time a new CMake is released? I don't want to review an endless stream of PRs bumping that version number for no benefit.

Unless I am misunderstanding this, setting `policy_max` to 3.21 implies that 6 different policies might be used (policies 3.16 through 3.21), depending on what version of CMake you are using. I think it would make sense to set `minimum_required` and `policy_max` to the same number to eliminate this ambiguity.

Someone using this project could add some code that depends on the policy for CMake 3.20 without realizing it; they will be in for a nasty surprise when someone else builds the project with CMake 3.19 and gets a different behavior. If both `minimum_required` and `policy_max` were set to 3.16, this kind of thing could not happen.

Additionally, I think it's worth adding a comment to describe what's happening here. Perhaps I am the only person in the world who will misunderstand how `policy_max` works, but somehow I doubt it.

src/sdl/CMakeLists.txt
ddalcino: > should I remove my comments?

Not yet. Without the comment, someone will want to change the code.

cmake/StandardProjectSettings.cmake
ddalcino: This is a good catch; #57 really should have used something more like `target_compile_options(project_options ...)` for this part instead.

However, I would prefer if you raised this issue in the [Issues](https://github.com/cpp-best-practices/cpp_starter_project/issues) section, rather than adding comments.

This is a very large PR, and fixing this problem is probably out of scope. If you were to fix this problem in a separate PR, I'm sure it would be merged much more quickly than this PR.

ddalcino:
> What is the problem with conan on my build host?

We can try to help you figure it out in the [Issues](https://github.com/cpp-best-practices/cpp_starter_project/issues) section; this does not belong in the discussion of your PR. If this project's CMake code needs to be changed to allow the project to build properly on your system, that needs to be part of a separate PR. It is more likely that we need to make changes to the README that describe a means of preventing whatever it is you're experiencing.

This PR is trying to accomplish at least 11 independent tasks by my count. Many of them are easy wins; if you had submitted them as separate PRs I'm sure they would have been merged by now.

src/sdl/main.cpp
```diff
@@ -1,18 +1,24 @@
...
+#include <SDL_error.h>    // for SDL_GetError
+#include <SDL_pixels.h>   // for SDL_MapRGB
+#include <SDL_surface.h>  // for SDL_FillRect, SDL_Surface
+#include <SDL_timer.h>    // for SDL_Delay
+#include <SDL_version.h>  // for SDL_version, SDL_GetVersion
+#include <SDL_video.h>    // for SDL_CreateWindow, SDL_DestroyWindow, SDL_Ge...
+
+#include <stdio.h>        // for fprintf, stderr
+#include <string.h>       // for strcmp
+#include <cstdlib>        // for NULL
```
ddalcino, line 15: IIRC, this file was copied directly from an SDL example; that's why it used the C version of `stdio.h`. 

If we are modernizing for C++, I think we really need to use the C++ versions of these headers: `<cstdio>` and `<cstring>`.

Also, if `NULL` is the only thing we need from `<cstdlib>`, we could remove that header and replace all the `NULL`s with `nullptr`. `nullptr` has been a part of the core language since C++11, and doesn't need to be defined by a macro in a library.

src/main.cpp
```diff
@@ -14,15 +22,15 @@ static constexpr auto USAGE =
...
+    Options:
           -h --help     Show this screen.
           --version     Show version.
           --speed=<kn>  Speed in knots [default: 10].
           --moored      Moored (anchored) mine.
           --drifting    Drifting mine.
 )";
 
-int main(int argc, const char **argv)
+int main([[maybe_unused]] int argc, [[maybe_unused]] char *argv[])
```
ddalcino, line 33: I think these variables are definitely being used. Was this change intended for `src/sdl/main.cpp`?

I like the change to `char *argv[]`, but I'm not sure why we lost the `const`.

ClausKlein: oh, my mistake! will be fixed

src/sdl/main.cpp
ClausKlein: I noticed that `include-what-you-use` recommends often the `c` header instead the `c++` header variant.

I think, I will split this MR into different parts next week. 

I am waiting for the cmakelib patch first.

aminya (changes requested):
I think most of the CMake changes are already integrated into project_options 
https://github.com/cpp-best-practices/project_options/pull/7
https://github.com/cpp-best-practices/project_options/pull/8
https://github.com/cpp-best-practices/project_options/pull/16#issuecomment-950383095

@ClausKlein please let me know if there is anything else remaining.

PR #173, opened 2021-10-19
Have built binaries be generated to 'build/bin/' instead of 'build/src/'

ShadowMitia:
I find it annoying/confusing that the binaries are generated in `builtd/src`. 

I'm suggesting a (very!) simple change to have it output to `build/bin` which makes more sense and should be easier to grasp for newcomers and beginners.

### DIFF
--- a/cmake/StandardProjectSettings.cmake
+++ b/cmake/StandardProjectSettings.cmake
@@ -40,3 +40,19 @@ else()
   message(STATUS "No colored compiler diagnostic set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
 endif()
 
+option(USE_OUTPUT_PATH "build all libaries and runtime files at lib and bin directory" YES)
+set(OUTPUT_LIB_PATH "/lib" CACHE STRING "Choose the output path of library targets.")
+set(OUTPUT_BIN_PATH "/bin" CACHE STRING "Choose the output path of binary targets.")
+if(USE_OUTPUT_PATH)
+  # -----------------------------------------------------------------------
+  # Where to put all the LIBRARY targets when built. This variable is used to initialize the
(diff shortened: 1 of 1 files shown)


aminya (changes requested):
This is not related to the output. It is the src code

ShadowMitia:
@aminya Sorry, I've updated my title/message to be more meaningful.

`add_subdirectory` accepts a second argument that defines what directory the binary will end up in. With the current setup that directory is named `src`, I'm suggesting to have it named `bin`.

ShadowMitia:
If you prefer, I don't mind either way. I just want to change that default value :grin:

ShadowMitia:
I've updated my proposal with @ClausKlein's proposal. While I was at it, I added the ability to set the custom directory name for binaries and libraries, keeping the default to "/bin" and "/lib".

aminya (changes requested):
I am not totally sold on this. It is changing a default setting of CMake which everyone is used to while not providing many benefits. CMake mirrors the exact structure of the source code, which is very intuitive. You can do this on your own fork, but I do not think this is a good idea for everyone.


Also, note that the actual build directory can be specified when you run configure command of Cmake.
```
cmake -S . -B ./bin
```

ShadowMitia:
Maybe I have another of seeing things true :grin: I dunno, it just feels simpler like that. Or maybe something more like Rust which has `target/debug` for debug builds (which here would give `build/target/debug`) ?

`cmake -S . -B ./bin` doesn't work because I would get `bin/src` and I'm thinking `build/bin`.

aminya:
You can propose an RFC here in the CMake repository.
https://gitlab.kitware.com/cmake/cmake/-/issues

ShadowMitia:
I don't need an RFC, there is already the `binary_dir` arg of `add_subdirectory`.

PR #175, opened 2021-11-27
Feature/cross ide cmake presets

Jason5480:
As discussed with @ddalcino in #161 It would be nice to have a cross IDE project configuration. Now this is available through CMakePresets.json file. A relative new CMake feature introduced in version 3.19 and it is already usable by VS2022 and VSCode  and CLion(not tested) in version 3.21. QtCreator will follow soon! 
I can verify that this work in Windows PC:
 - using Visual Studio 2022 and msvc compiler in Local machine
 - using Visual Studio 2022 and a WSL Ubuntu18.04, Ubuntu20.04 with both gcc and clang
 - using VS Code and msvc compiler in Local machine
 - using VS Code and the Docker image provided in this repo Ubuntu18.04 both with gcc and clang

It would be nice if a Linux user with CLion and VSCode installe can also verify that this work as expected (@lefticus :))
- [x]  In his Local machine using both gcc and clang in VS Code
- [x]  In his Local machine using both gcc anf clang in CLion (probably more vendor specific options are needed as described [here](https://www.jetbrains.com/help/clion/cmake-presets.html).
- [x] using VSCode and the Docker image gcc and clang
- [x] any other inception concept

Also it would be nice to have some feedback (input and thoughts) from @midnightexigent and  @shreyasbharath  who implemented #60 and #99 and have experience with VSCode and Docker. 

EDIT:
Dockerfile is updated to also support ubuntu focal and the most recent gcc and clang versions (always configurable when build the image). New ${VARIANT} equal to [bionic, focal] should do the job like GCC_VER, LLVM_VER and USE_CLANG


Jason5480:
Also I have to mention that it was not necessary to increase the minimum CMake version for IDE. VS and VSCode can work with that. However, if we want consistency with the CI we have to use the `cmake --build --preset=...` and this need the newer CMake version! FYI (@ddalcino)

CMakePresets.json
```diff
@@ -0,0 +1,384 @@
+{
+    "version": 3,
```
ddalcino, line 2: I am trying this right now on Linux with CLion, and so far I have observed that [CLion 2021.2.3 supports only version 2 Presets and only buildPresets](https://www.jetbrains.com/help/clion/cmake-presets.html).

When version is set to 3, as it is here, none of the presets are loaded. I am not fully convinced that this is a bad thing; CLion works well enough with this project without the presets. However, if we want this to work with CLion right now, this version number needs to come down to 2.

I would be surprised if CLion does not provide support for version 3 of CMakePresets.json at some point in the future, so there may be some value to leaving the version number as-is.

The CMake [docs](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) describe the differences between versions 2 and 3; I did not see anything in the docs that made it look like we really need to use version 3.

CMakePresets.json
```diff
@@ -0,0 +1,384 @@
...
+                    ],
+                    "intelliSenseMode": "windows-clang-x64"
+                }
+            }
+        },
+        {
+            "name": "linux-gcc-debug",
+            "displayName": "gcc Debug",
+            "description": "Target Linux with the gcc compiler, debug build type",
+            "generator": "Ninja",
```
ddalcino, line 111: Should we specify `ninja` here for the generator? AFAIK `make` tends to be installed by default on most Linux/Mac systems with any kind of C++ toolchain installed, and `ninja` typically is not. 

If we are going to specify `ninja` here, we will need additional documentation on how to install it.

Is `ninja` bundled with MSVC? I am curious about how `ninja` got into all the Windows builds.

Note that if we are using CMakePresets version 3, the `generator` field "may be omitted to fall back to regular generator discovery procedure." (docs: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset)

CMakePresets.json
```diff
@@ -0,0 +1,384 @@
...
+                "-v"
+            ],
+            "configurePreset": "linux-gcc-release"
+        },
+        {
+            "name": "default-build-linux-clang-debug",
+            "displayName": "Default",
+            "description": "Vanilla build",
+            "configurePreset": "linux-clang-debug"
+        },
```
ddalcino, line 292: For this build, on Linux/CLion, I am getting a repeat of issue #40: CMake thinks it's using clang, but conan thinks it's using the default c++ compiler, which is g++. The fix proposed in #122 resolves this issue.

Whatever went wrong here for me might be a consequence of the way CLion implements support for CMakePresets: As discussed above, CLion only supports `buildPresets`.

I am a little confused by what the [CLion docs](https://www.jetbrains.com/help/clion/cmake-presets.html) mean when they say CLion "only supports `buildPresets`. When using this preset, CMake knows that it's supposed to use `clang`, `clang++`, and `ninja`. These tools are not listed in `buildPresets`, but they are listed in the `configurePreset` item specified here. Empirically, it appears that CLion supports `configurePresets` as well as `buildPresets`. Perhaps the docs are wrong?

ddalcino (changes requested):
Can we get some additions to the documentation that describe how this is meant to be used? You don't have to make it too elaborate; maybe a quick note on how to use the `--preset` flag with `cmake` on the command line, and a note/link to a guide on how to use presets with VS/VSCode would be sufficient. We can worry about CLion/QtCreator later, after they implement full support for presets.

CMakePresets.json
Jason5480: Lets create an issue to re-test this CMake presets implementation when CLion and QtCreator have a completed their support for this feature. Those should come very soon.

CMakePresets.json
Jason5480: Ninja is the default generator for the msvc toolchain so yes it is working out of the box in windows. Now, since it has big performance advantage over make I thought it would be a good choice to use it  in Linux as well. To be honest, I also made some improvements to the Dockerfile and I included the `ninja-build` so it can work in my windows machine using docker. I will push that changes so please when you find some time see those as well.

CMakePresets.json
Jason5480: Also, I noticed that clang for windows in not able to build docopt.. It should be a conan bug?? (https://github.com/microsoft/vscode-cmake-tools/issues/2090) I don't know. I hope #122 fix that too. For now on windows local machine I can use only msvc toolchain.

Jason5480:
Yes I will update the documentation too. @ddalcino Thank you for the review!

CMakePresets.json
Jason5480: @ddalcino new commits are pushed with several improvements I did. One of them was to add ninja-build

.devcontainer/Dockerfile
```diff
@@ -5,32 +5,42 @@ FROM ubuntu:${VARIANT}
 # Restate the variant to use it later on in the llvm and cmake installations
 ARG VARIANT
 
-# Install packages available from standard repos
+# Install nessesary packages available from standard repos
```
ddalcino, line 8:
```suggestion
# Install necessary packages available from standard repos
```

.devcontainer/Dockerfile
```diff
@@ -53,6 +76,12 @@ RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/nul
     apt-get update -qq && export DEBIAN_FRONTEND=noninteractive && \
     apt-get install -y --no-install-recommends cmake cmake-curses-gui
 
+# Install optional dependecies
+RUN apt-get update -qq && export DEBIAN_FRONTEND=noninteractive && \
+    apt-get install -y --no-install-recommends \
+        doxygen ccache cppcheck \
+        neovim emacs nano
```
ddalcino, line 83: I appreciate the effort to split packages into 'necessary' vs 'optional', but it looks like you lost the `graphviz` packages in this change. `graphviz` is important for `doxygen`; we need that.

.devcontainer/Dockerfile
```diff
@@ -44,6 +54,19 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - 2>/dev/
...
+RUN update-alternatives --install /usr/bin/clang++ clang++ $(which clang++-${LLVM_VER}) 100
+
+# Allow the user to set compiler defaults
+ARG USE_CLANG
+# if --build-arg USE_CLANG=1, set CC to 'clang' or set to null otherwise.
+ENV CC=${USE_CLANG:+"clang"}
+ENV CXX=${USE_CLANG:+"clang++"}
+# if CC is null, set it to 'gcc' (or leave as is otherwise).
+ENV CC=${CC:-"gcc"}
+ENV CXX=${CXX:-"g++"}
```
ddalcino, line 68: When I wrote this, I purposefully placed this argument at the very end of the file so that users can turn `USE_CLANG` on or off without needing to rebuild much of anything at all. Now that it has been moved up, if you change `USE_CLANG` and rebuild the image, the cache for `cmake` and all of the optional dependencies is invalidated, and will be installed again.

I prefer this section at the end of the file.

CMakePresets.json
ddalcino: CLion 2021.3.1 was released Dec. 1, and the docs say that it supports version 3. I haven't tested it yet, but at this point I don't think there's any reason to drop the version down to 2.

I'm not sure what's going on with QtCreator and CMake Presets. [The issue QTCREATORBUG-24555](https://bugreports.qt.io/browse/QTCREATORBUG-24555) was created in August 2020, and no progress has been noted in the bug tracker. 

I don't think we need to make sure that every IDE plays nice with this project. There are too many IDEs! I think VS Code is enough to worry about.

.gitignore
```diff
@@ -1,18 +1,19 @@
...
+**/out/
+**/cmake-build-*/
 
 # IDE files
-.vs/
-.idea/
-.vscode/
+**/.vs/
+**/.idea/
+**/.vscode/
```
ddalcino, line 9: I am not clear on why this change is being made. The commit message just says "ignore folders recursively". Is there a reason for adding this? Are there folders that VS Code adds to your repo all over your project, rather than just the root level?

Keep in mind that this is a general-purpose best-practices template project. Users are expected to add new items to the `.gitignore` as needed. The default `.gitignore` is meant to provide sane defaults that anyone can use, with minimal impact on anyone that hasn't read it yet. With this change, I could add a `src/util/build/` directory full of cpp files, and if I do not carefully read my `.gitignore`, I would not be able to commit anything in the directory. That would be a nasty surprise.

CMakePresets.json
Jason5480: Yes  I agree if we follow the de facto standard all the IDE will follow sooner or later!

.devcontainer/Dockerfile
Jason5480: oups nice catch!

.devcontainer/Dockerfile
Jason5480: Hmm yes it is better at the end

.gitignore
Jason5480: I accidentally opened the src/qt with visual studio using the open folder functionality and the VS created the .vs folder under src/qt/.vs. It was just an accident. I know that only the root folder should be the entry point since the main CMakeLists is there. I can revert that I do not consider this change important

.devcontainer/Dockerfile
```diff
@@ -84,17 +108,8 @@ ENV CXX=${USE_CLANG:+"clang++"}
...
-ENV CONAN_SYSREQUIRES_SUDO 0
-# Some packages request that Conan use the system package manager to install
-# a few dependencies. This flag allows Conan to proceed with these installations;
-# leaving this flag undefined can cause some installation failures.
-ENV CONAN_SYSREQUIRES_MODE enabled
-
 # Include project
-ADD . /starter_project
-WORKDIR /starter_project
+#ADD . /workspaces/cpp_starter_project
```
Jason5480, line 112: @ddalcino The more I see this copy the less best practice it seems to me. This is because if you make changes in the code inside the container they will be lost when you delete that container. The way to go here is to create a volume as described in the readme file (if you don't use VSCode that does that automatically). So, I suggest to remove that entirely to avoid loosing code changes. Also, it takes time when you build the container. What do you think?

ddalcino: I think it depends on how you’re using the Dockerfile. If you’re developing locally, then it doesn’t make any sense to copy the source folder. If it’s part of your CI/CD pipeline, it might be easier to copy the source code than to mount a directory. Everyone’s needs are different.

Jason5480: > 

Shouldn't the CI clone the repo and check it out to the correct branch before everything else? Well I am not qualified by any means when it comes to the CI jobs.. So should I un-comment those two lines?

ddalcino: Those lines should definitely be commented out, if not removed entirely. I think that the Dockerfile should work properly with VSCode out-of-the-box. If someone is using the Dockerfile in their CI, they will need to make several modifications to get everything working the way they want it to. I don't think that we can predict what those modifications should be.

Honestly, I do not know if we should keep these two lines as comments, or if we should remove them. As comments, they do not bother me. I think that what you have for this file is just fine.

.gitignore
```diff
@@ -13,6 +13,8 @@ cmake-build-*/
 !.vscode/extensions.json
 *.swp
 *~
+**/_ReSharper*
```
ddalcino, line 16: I'm not sure how this line got in here. Is this necessary there a need for this, or did you intend to remove the `**/` in 4f9676672e512c9fdff2adaf55b9d2b059bec8e0?

.gitattributes
```diff
@@ -0,0 +1,65 @@
...
+*.pdf filter=lfs diff=lfs merge=lfs -text
+*.png filter=lfs diff=lfs merge=lfs -text
+*.psd filter=lfs diff=lfs merge=lfs -text
+*.webp filter=lfs diff=lfs merge=lfs -text
+
+# Fonts
+*.woff2 filter=lfs diff=lfs merge=lfs -text
+
+# Other
+*.exe filter=lfs diff=lfs merge=lfs -text
```
ddalcino, line 65: I don't mean to nitpick, but I am not sure what adding a `.gitattributes` file has to do with this feature. I think that this could easily be a separate PR. Is this necessary for CMakePresets/VSCode support?

README.md
```diff
@@ -338,6 +338,10 @@ With Cmake directly:
     cmake -S . -B ./build
 
 Cmake will automatically create the `./build` folder if it does not exist, and it wil configure the project.
+However, if you have CMake version 3.21+ it is recomended to use one of the presets that are listed in the CmakePresets.json file.
```
ddalcino, line 341:
```suggestion
However, if you have CMake version 3.21+ it is recommended to use one of the presets that are listed in the CmakePresets.json file.
```

.gitignore
Jason5480: Yes I wanted to remove the `**/` in this commit. Well Resharper is a very popular plugin from Jetbrains for VS that gives CLion capabilities to the IDE. This is the folder that Resharper writes its chase files.

.gitattributes
Jason5480: There was a problem with the line endings. When I was developing inside the Docker the VSCode used LF because it was in a Linux OS but when I was developing in VSCode and native windows it was using CRLF. So git diff showed that the whole file had changed.

.gitignore
ddalcino:
```suggestion
_ReSharper*
```

README.md
```diff
@@ -497,26 +501,30 @@ See [libFuzzer Tutorial](https://github.com/google/fuzzing/blob/master/tutorial/
...
 ```bash
 docker build --tag=my_project:latest .
 docker run -it my_project:latest
 ```
 
-This command will put you in a `bash` session in a Ubuntu 18.04 Docker container,
+This command will put you in a `bash` session in a Ubuntu 20.04 Docker container,
 with all of the tools listed in the [Dependencies](#dependencies) section already installed.
-Additionally, you will have `g++-10` and `clang++-11` installed as the default
+Additionally, you will have `g++-11` and `clang++-12` installed as the default
```
ddalcino, line 517:
```suggestion
Additionally, you will have `g++-11` and `clang++-13` installed as the default
```

README.md
```diff
@@ -546,7 +554,7 @@ You can configure and build [as directed above](#build) using these commands:
 /starter_project# cmake --build ./build
 ```
 
-You can configure and build using `clang-11`, without rebuilding the container,
+You can configure and build using `clang-12`, without rebuilding the container,
```
ddalcino, line 557:
```suggestion
You can configure and build using `clang-13`, without rebuilding the container,
```

ddalcino (approved):
Looks good to me.

Note that all changes to `conanfile.py` and `cmake/Conan.cmake` will be overwritten by #169. I think this is acceptable, because those changes are not the really the point of this PR. Changes to `cmake/Conan.cmake` may be submitted as PRs to https://github.com/aminya/ProjectOptions/

.gitattributes
```diff
@@ -0,0 +1,65 @@
...
+
+# Force batch scripts to always use CRLF line endings so that if a repo is accessed
+# in Windows via a file share from Linux, the scripts will work.
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
+
+# Force bash scripts to always use LF line endings so that if a repo is accessed
+# in Unix via a file share from Windows, the scripts will work.
+*.sh text eol=lf
+
```
aminya, line 37: This whole section can be written much more succinctly. There is no need to declare the extensions one by one
```
# line endings
* text=auto eol=lf
*.{cmd,[cC][mM][dD]} text eol=crlf
*.{bat,[bB][aA][tT]} text eol=crlf
*.{vcxproj,vcxproj.filters} text eol=crlf
```

aminya (changes requested):

.gitattributes
```diff
@@ -0,0 +1,65 @@
...
+###############################
+# Git Large File System (LFS) #
```
aminya, line 39: Git LFS is not free, and there is a size limit on GitHub. Not sure if we should use it by default for all of these extensions.

aminya (changes requested):

README.md
```diff
@@ -497,26 +501,30 @@ See [libFuzzer Tutorial](https://github.com/google/fuzzing/blob/master/tutorial/
...
+in your terminal, when the Dockerfile is inside the .devconatiner directory:
+
+```bash
+cd .devcontainer/
+```
 
 ```bash
 docker build --tag=my_project:latest .
 docker run -it my_project:latest
 ```
```
aminya, line 513: Instead of cd you can specify the docker file name.
```bash
docker build -f ./.devcontainer/Dockerfile --tag=my_project:latest .
docker run -it my_project:latest
```

aminya (changes requested):

CMakePresets.json
```diff
@@ -0,0 +1,396 @@
...
+        {
+            "name": "verbose-build-windows-msvc-debug",
+            "displayName": "Verbose",
+            "description": "Passes -v to Ninja",
+            "nativeToolOptions": [
+                "-v"
+            ],
+            "configurePreset": "windows-msvc-debug"
+        },
```
aminya, line 223: I don't think adding a separate entry for --verbose is a good idea. If someone runs into such low-level issues (very rare), they would have an easier time reading `build.ninja` file.

```suggestion
```

CMakePresets.json
```diff
@@ -0,0 +1,396 @@
...
+                "CMAKE_C_COMPILER": "cl",
+                "CMAKE_CXX_COMPILER": "cl"
+            },
+            "vendor": {
+                "microsoft.com/VisualStudioSettings/CMake/1.0": {
+                    "hostOS": [
+                        "Windows"
+                    ]
+                }
+            }
```
aminya, line 31: Why do you specify the hostOS manually?

aminya (changes requested):

aminya:
> Note that all changes to `conanfile.py` and `cmake/Conan.cmake` will be overwritten by #169. I think this is acceptable because those changes are not the really the point of this PR. Changes to `cmake/Conan.cmake` may be submitted as PRs to [aminya/ProjectOptions](https://github.com/aminya/ProjectOptions/)

Pull requests are welcome! I can merge the changes myself if you cannot make a PR.

CMakePresets.json
Jason5480: This is a vendor specific option for VS and VSCode. This tells the IDE to hide that preset if the host is not Windows. To be honest I didn't check what it does if it is not defined. My guess is that you will be able to select that preset through you IDE menus when you are in Linux.

CMakePresets.json
Jason5480: Yes this is why I also have the "Default" build preset for windows-msvc-debug and all other configure-presets. Default will just run ninja, but if something is fishy the user can get more info by selecting this build-preset (or any other verbose)

README.md
Jason5480: yhep I will change that!

CMakePresets.json
aminya: To reduce the number of presets, we can remove the ones that only add a --verbose option (rare usecase). If someone runs into a problem, they can go into their cmake and modify it manually.

.gitattributes
Jason5480: https://github.com/git-lfs/git-lfs/blob/main/LICENSE.md

CMakePresets.json
Jason5480: who want to have less options? The fact that there is a verbose option does not mean that you have to build it that way. Well I can remove that if you insist but when you set up a new project it might be useful to have it out of the box.

.gitattributes
ddalcino: > https://github.com/git-lfs/git-lfs/blob/main/LICENSE.md

https://docs.github.com/en/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage#about-billing-for-git-large-file-storage

1GB is free

aminya: Yeah, the software is free but the storage is not.

CMakePresets.json
aminya: I think it is desirable to keep the number of presets manageable. Imaging if someone wants to modify a build preset. Now, they have to do it in two places. For the regular one and the verbose one.

Jason5480: Fair point! As I can see default/empty build preset already exist, there is no need to define it 
image
Shall I remove the user defined Default build presets too?

.gitattributes
Jason5480: Sorry, I am used to self-hosted solutions. Well if you have big binary assets for your application you cannot let them getting diffed  like a text it is a big git slowdown with out any meaningful outcome.

CMakePresets.json
aminya: If that translates to other IDEs too, that's probably a good idea.

Jason5480: Visual Studio has the same behavior with the VSCode when it comes to the default build preset. It makes no sense to define a preset if you do not alter the option arguments. You could just not pass any preset in the cli! I guess this is what IDEs does too. So if we are going to keep only one build preset for demonstration reasons I would only keep the verbose since the default can be omitted.  Otherwise, I can erase all the buildPresets and if someone want to pass something special he can search a little bit more.

Jason5480: @aminya if you have any useful build option to suggest that a build preset could possibly pass to the ninja it would be great.

aminya: I prefer that we remove the default presets.

Jason5480: OK I made that change. We should be good to go now

Jason5480: @aminya is this OK in the latest commit? I think we ready to go if you do not have anything else to propose. The default and verbose build presets were deleted as discussed above.

Jason5480: @aminya I tried to re-use configurations between different presets as much as possible, so there is one place to define a single flag. Now I have some common presets and other presets inherit from them. I think this is a good addition to minimize the maintenance cost of the CMakePresets.json file.

.gitattributes
Jason5480: Given that users will upload only assets that are necessary for their applications and not binaries the free 1G should be more than enough. Also there is the option to package and version your assets using conan, but this is a very specific use case, (probably for an other MR) However, implementing something like this I am afraid push too far for a multi purpose template project.

Jason5480:
Hello guys @aminya @ddalcino Merry Christmas! I did the changes you requested can you please verify and move this forward? Thank you and happy holidays!

.gitignore
```diff
@@ -7,12 +7,10 @@ cmake-build-*/
 .vs/
 .idea/
 .vscode/
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
```
aminya, line 13: Please revert this change. Many people commit their workspace and project settings so others on the team have access to the same settings and debug commands

.gitignore
```diff
@@ -7,12 +7,10 @@ cmake-build-*/
 .vs/
 .idea/
 .vscode/
-!.vscode/settings.json
```
aminya, line 10: Please revert this change. Many people commit their workspace and project settings so others on the team have access to the same settings and debug commands

README.md
```diff
@@ -339,6 +339,11 @@ With Cmake directly:
 
 Cmake will automatically create the `./build` folder if it does not exist, and it wil configure the project.
 
+However, if you have CMake version 3.21+ it is recommended to use one of the configuration presets that are listed in the CmakePresets.json file.
+
+    cmake . --preset <configure-preset>
+	cmake --build
```
aminya, line 345:
```suggestion
    cmake . --preset <configure-preset>
    cmake --build
```

README.md
```diff
@@ -339,6 +339,11 @@ With Cmake directly:
 
 Cmake will automatically create the `./build` folder if it does not exist, and it wil configure the project.
 
+However, if you have CMake version 3.21+ it is recommended to use one of the configuration presets that are listed in the CmakePresets.json file.
```
aminya, line 342:
```suggestion
Instead, if you have CMake version 3.21+, you can use one of the configuration presets that are listed in the CmakePresets.json file.
```

aminya (changes requested):
Some final changes

.gitattributes
aminya: I still want this to be reverted. You can add a link to a Gist explaining this instead. Repos with Git LFS are also slower to clone

Jason5480: OK I  removed that part of the .gitattributes. I still believe that setting lfs from the early stages of the project to decrease the overall repo size on the client worth the trade off.

Jason5480: https://stackoverflow.com/questions/35575400/what-is-the-advantage-of-git-lfs

.devcontainer/.dockerignore
```diff
@@ -0,0 +1,34 @@
+# Build directories and binary files
```
aminya, line 1: I am not familiar with .dockerignore. What is its purpose? Why does not it use the gitignore file?

.gitattributes
aminya: You could leave it there but just commented out. If someone has very big files, it is really useful to use LFS.

I totally agree that LFS is useful, and solves one of the issues of Git. My main problem with it is that it is not free on GitHub. 
I used to use it on GitHub for free, and I constantly hit the limit. The limit prevented any clones or pushes! That's why I prefer this to be opt-in rather than by default.

aminya (approved):

aminya:
Let's wait for @lefticus input on this before merging this.

.devcontainer/.dockerignore
Jason5480: At the very end of the docker image creation when we add the workspace we ignore the files listed in that file. Well I don't know the rationale why it is not using the .gitignore file. For dev-containers it seems logical to me but I as discussed [here](https://stackoverflow.com/questions/58707272/should-dockerignore-typically-be-a-superset-of-gitignore) there are many other use cases for docker out there.

Jason5480:
> Let's wait for the @lefticus input on this before merging this.

OK lets wait @lefticus input on this. Btw I have one more point for discussion. When I started this issue, I used "Ninja" generator  for the presets as it is the fastest and more general generator that work with most of the IDEs. However, there is a "new player" called "Ninja Multi-Config" designed to work better with IDEs and let the developer to change from release to debug build faster. Also using that generator build presets have an actual meaning by setting the configuration to (Debug, Release). If that change is approved for development I will open a different merge request since this one already contains many changes that are hard to follow.

lefticus:
I have no real input on the docker files, and as far as I understand it, the cmake-presets are optional as to whether they are utilized, is that correct. @Jason5480 ?

.devcontainer/.dockerignore
aminya: Is there a way to automate copying `.gitignore` and `Dockerfile` to these .devcontainer files? Not everyone uses VsCode

Jason5480:
> I have no real input on the docker files, and as far as I understand it, the cmake-presets are optional as to whether they are utilized, is that correct. @Jason5480 ?

Yes, you still can pass the arguments one by one in the cli nothing changed here. Presets are a nice way to bundle all those arguments together, give them a descriptive name and share them across a team. You may have presets for the cmake configuration phase, build phase and test phase. Optionally, you can pass the `--preset=<name>` in the cli and the cmake will pass the arguments defined in the corresponding preset.
 The great benefit I can see is that (in the future) all IDEs will have a common configuration project file, already MS replaced the old CMakeSettings.json and moved to the standard CMakePresets.json and CMakeUserPresets.json. Thus, Visual Studio and VSCode have full presets integration. As far as I know, CLion has partial integration (for build) and QtCreator have scheduled it in its roadmap.
Now, Docker is very convenient for those who have not native Linux and they don't want to spend much time to set up a VM from scratch. As for current changes on Dockerfile they are a great improvement since the user can select the Linux distro, gcc and clang most recent versions and also has a seamless integration with VSCode (but also can be used manually). That's all

lefticus (approved):

PR #183, opened 2022-01-23
ci: use setup-cpp to set up all compilers and tools for all the CI services (GitHub, Gitlab, etc.)

aminya:
https://github.com/aminya/setup-cpp

use setup-cpp to set up all compilers and tools for all the CI services (GitHub, Gitlab, etc.). This includes different compilers and platforms:

image


lefticus (approved):
I'm in favor of more automated setup, and more CI tests

lefticus:
should we drop appveyor? Does it make sense to keep maintaining it when github actions are so easy to use?

aminya:
> should we drop appveyor? Does it make sense to keep maintaining it when github actions are so easy to use?

I agree. I removed it. You should also disable its app from the settings.

ddalcino (commented):
I guess my approval doesn’t matter now, but this looks really good to me. Nice work, @aminya!

PR #190, opened 2022-02-05
Test using setup-python directly for common packages

lefticus:
* `gcovr`, `conan`, `ninja`, `cmake` are all fully supported with pip install on all 3 platforms
 * using setup-python and pip3 directly removes one variable from the build process
 * using pip3 directly to install all 3 provides consistency across platforms

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,22 +101,26 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ matrix.compiler }}-${{matrix.build_type}}-${{matrix.generator}}-${{matrix.developer_mode}}
 
+      - uses: actions/setup-python@v2
+
+      - run: pip3 install cmake ninja conan gcovr
+
       - name: Setup Cpp
         uses: aminya/setup-cpp@v1
         with:
           compiler: ${{ matrix.compiler }}
           vcvarsall: ${{ contains(matrix.os, 'windows' )}}
 
-          cmake: true
-          ninja: true
-          conan: true
+          cmake: false
+          ninja: false
+          conan: false
           vcpkg: false
           ccache: true
           clangtidy: true
 
           cppcheck: true
 
-          gcovr: true
+          gcovr: false
           opencppcoverage: true
 
       - name: Configure CMake


lefticus:
This partially a workaround for https://github.com/aminya/setup-cpp/issues/27 but also I think it is not a bad idea, as mentioned in the commit message, it gives consistency across all platforms for the versions and mechanism that common tools are installed.

ddalcino:
Just for clarification: Is this meant to be permanent, or just a temporary fix until https://github.com/aminya/setup-cpp/issues/27 is fixed?

ddalcino (approved):

lefticus:
> Just for clarification: Is this meant to be permanent, or just a temporary fix until [aminya/setup-cpp#27](https://github.com/aminya/setup-cpp/issues/27) is fixed?

In my opinion, it would be permanent. I see it as one less potential issue while maintaining this project, and for our users in the future. I don't see any reason to use a mix of chocolatey, package managers, etc, when python gives us the same versions of the same packages across all platforms.

aminya (changes requested):
It is not setup-cpp's fault that the pre-installed Python broke on Windows images. In fact setup-cpp does not call "setup-python" until it actually needed. I do not approve of this, as this is like discarding all the things it was considered in setup-cpp and trying to re-implement them in bash.

.github/workflows/ci.yml
```diff
@@ -101,22 +101,26 @@ jobs:
...
         uses: aminya/setup-cpp@v1
         with:
           compiler: ${{ matrix.compiler }}
           vcvarsall: ${{ contains(matrix.os, 'windows' )}}
 
-          cmake: true
-          ninja: true
-          conan: true
+          cmake: false
+          ninja: false
```
aminya, line 115: Cmake and Ninja do not need python. Why should they be installed with pip3?

aminya, line 115: Cmake and Ninja do not need python. Why should they be installed with pip3?

.github/workflows/ci.yml
```diff
@@ -101,22 +101,26 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ matrix.compiler }}-${{matrix.build_type}}-${{matrix.generator}}-${{matrix.developer_mode}}
 
+      - uses: actions/setup-python@v2
+
+      - run: pip3 install cmake ninja conan gcovr
```
aminya, line 106: This is incomplete. It does not put the paths that conan and gcovr are called from on the path. Have you tried this inside Docker or locally?? 

https://github.com/aminya/setup-cpp/blob/94827f017e0e4445837e4cc099daf7ce849a059c/src/utils/setup/setupPipPack.ts#L35-L56

aminya (changes requested):

.github/workflows/ci.yml
aminya, line 106: On the GitHub images, it happens that `pip3` is on the path. But a lot of the time, it is actually called `pip`, which means it is needed to check if `pip` is version 3 or above. 

https://github.com/aminya/setup-cpp/blob/94827f017e0e4445837e4cc099daf7ce849a059c/src/utils/setup/setupPipPack.ts#L19-L29

aminya (changes requested):

.github/workflows/ci.yml
lefticus: > On the GitHub images, it happens that `pip3` is on the path. But a lot of the time, it is actually called `pip`, which means it is needed to check if `pip` has version 3 or above.
> 

But isn't this yml file *only* used on github images?

It's specifically in `.github/workflows/ci.yml`, so shouldn't that be the primary concern here?

.github/workflows/ci.yml
lefticus: Because it gives a simple consistent command that works on all 3 platforms that are supported on github. Since we are working specifically on a github workflow here.

lefticus:
> It is not setup-cpp's fault that the pre-installed Python broke on Windows images. In fact setup-cpp does not call "setup-python" until it actually needed. I do not approve of this, as this is like discarding all the things it was considered in setup-cpp and trying to re-implement them in bash.

It is not discarding everything setup-cpp is used for. Setup-cpp still does all of the hard work of finding and installing compilers across compilers.

aminya (changes requested):
The issue is fixed in setup-cpp. The pre-installed pip executable is broken on Windows images. The workaround is to use `python -m pip install package`

https://github.com/aminya/setup-cpp/issues/27#issuecomment-1030691291

I don't think this is needed anymore

aminya:
> > It is not setup-cpp's fault that the pre-installed Python broke on Windows images. In fact setup-cpp does not call "setup-python" until it actually needed. I do not approve of this, as this is like discarding all the things it was considered in setup-cpp and trying to re-implement them in bash.
> 
> It is not discarding everything setup-cpp is used for. Setup-cpp still does all of the hard work of finding and installing compilers across compilers.

Here I specifically meant the pip installations. As I mentioned in the above comments, it is needed to check if:
- python paths are added to PATH
- if pip is up to date

.github/workflows/ci.yml
lefticus: > > > It is not setup-cpp's fault that the pre-installed Python broke on Windows images. In fact setup-cpp does not call "setup-python" until it actually needed. I do not approve of this, as this is like discarding all the things it was considered in setup-cpp and trying to re-implement them in bash.
> > 
> > 
> > It is not discarding everything setup-cpp is used for. Setup-cpp still does all of the hard work of finding and installing compilers across compilers.
> 
> Here I specifically meant the pip installations. As I mentioned in the above comments, it is needed to check if:
> 
>     * python paths are added to PATH on all platforms
> 
>     * if pip is installed and is up to date

It's good that setup-cpp is not fixed, but I think it's worth mentioning that all of this worked without any other path manipulations. All tests passed and even properly posted coverage reports.

ddalcino: I think that what @aminya’s not saying here is that setup-cpp works just as well on a self-hosted runner. The setup-Python action takes a little more work for that. Some users depend on being able to build their projects with a self-hosted runner; maybe we should try to support them.

ddalcino: We could add additional code here to make self-hosted runners work; there are directions on how to do that here: https://github.com/actions/setup-python.

Or, we could just use setup-cpp.

lefticus:
I'm just going to close and remove this PR, since setup-cpp's python integration is working again.

aminya:
Worth mentioning that the pip issue is a serious problem in their launchers. It is affecting many people on Windows. 
See the progress here:
https://github.com/pypa/pip/issues/10875

PR #191, opened 2022-02-06
Fix coverage reporting

lefticus:
* Need to make sure opt_enable_coverage is set when appropriate (only for Debug builds, since it overrides optimization settings)
 * Remove explicit sources dir from windows coverage, so that it still works when template project is used
 * Make sure gcov version/executable is set correctly
 * Add simple test of main executable, to show its coverage also

Note: these issues were discovered while using a recent project made from the template. I think most of these issues are long standing. Previously `ENABLE_COVERAGE` was enabled by default, so it happened to work, but Release builds were not doing what we thought they were doing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,16 +53,16 @@ jobs:
           # Add appropriate variables for gcov version required. This will intentionally break
           # if you try to use a compiler that does not have gcov set
           - compiler: gcc
-            gcov: gcov
+            gcov_executable: gcov
           - compiler: llvm
-            gcov: "llvm-cov gcov"
+            gcov_executable: "llvm-cov gcov"
 
           # This exists solely to make sure a non-multiconfig build works
           - os: ubuntu-20.04
             compiler: gcc
             generator: "Unix Makefiles"
             build_type: Debug
-            gcov: gcov
+            gcov_executable: gcov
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -50,11 +50,19 @@ jobs:
             compiler: gcc
 
         include:
+          # Add appropriate variables for gcov version required. This will intentionally break
+          # if you try to use a compiler that does not have gcov set
+          - compiler: gcc
+            gcov: gcov
```
aminya, line 56: Should we choose a better name for this variable? I am afraid it is confused with the gcovr input passed to setup-cpp

aminya (changes requested):

test/CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ add_library(catch_main OBJECT catch_main.cpp)
 target_link_libraries(catch_main PUBLIC Catch2::Catch2)
 target_link_libraries(catch_main PRIVATE project_options)
 
+add_test(NAME cli.has_help COMMAND intro --help)
```
aminya, line 10: A description for this test would be useful

aminya (approved):

ddalcino (approved):

PR #193, opened 2022-02-10
Check MacOS-latest too, since apple is always breaking things

lefticus:


ddalcino (approved):
Yes, this is a good idea.

At some point, we might want to look at universal builds on Mac (I think that’s what it’s called). This would build binaries for both x86_64 and arm. As more users move to apple silicon, I think our users may want this.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           - ubuntu-20.04
           - macos-10.15
```
aminya, line 32: Testing two macos builds is unnecessary. This just adds more pressure on GitHub's CI servers.

```suggestion
```

PR #195, opened 2022-02-11
Test the possibility of using problem matchers for errors/warnings

lefticus:


.github/workflows/ci.yml
```diff
@@ -133,6 +133,11 @@ jobs:
         run: |
           cmake -S . -B ./build -G "${{matrix.generator}}" -DCMAKE_BUILD_TYPE:STRING=${{matrix.build_type}} -DENABLE_DEVELOPER_MODE:BOOL=${{matrix.developer_mode}} -DOPT_ENABLE_COVERAGE:BOOL=${{ matrix.build_type == 'Debug' }}
 
+      - name: GCC Problem Matcher
+        uses: ammaraskar/gcc-problem-matcher@0.1
+      - name: MSVC Problem Matcher
+        uses: ammaraskar/msvc-problem-matcher@0.1
+
```
aminya, line 140: Oh, cool. I think I can integrate this into setup-cpp!

aminya: In progress: 
https://github.com/aminya/setup-cpp/pull/35

aminya: Done
```suggestion
```

aminya:
Does clang uses the same report style as gcc?
Related to https://github.com/aminya/setup-cpp/issues/34

lefticus:
> Does clang uses the same report style as gcc? Related to [aminya/setup-cpp#34](https://github.com/aminya/setup-cpp/issues/34)

It is very similar, but isn't matching.

I'm pretty confused, TBH

clang, which doesn't match:

https://github.com/cpp-best-practices/cpp_starter_project/runs/5164418830?check_suite_focus=true#step:6:20

```
/home/runner/work/cpp_starter_project/cpp_starter_project/src/main.cpp:47:13: warning: variable 'j' is uninitialized when used here [-Wuninitialized]
    int l = j + k;
```	

clang-tidy, which does match:

https://github.com/cpp-best-practices/cpp_starter_project/runs/5164419289?check_suite_focus=true#step:6:23

```	
/home/runner/work/cpp_starter_project/cpp_starter_project/src/main.cpp:43:10: error: 42 is a magic number; consider replacing it with a named constant [cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers,-warnings-as-errors]
```

gcc, which does match:

https://github.com/cpp-best-practices/cpp_starter_project/runs/5164419322?check_suite_focus=true#step:6:16

```
/Users/runner/work/cpp_starter_project/cpp_starter_project/src/main.cpp:47:15: warning: 'j' may be used uninitialized [-Wmaybe-uninitialized]
   47 |     int l = j + k;
      |             ~~^~~
```

lefticus:
I just ran them through a regex tester and don't see what it's not matching

aminya:
The gcc matcher is only enabled if the compiler is gcc. If the regex is the same as clang, I can add the same thing to clang and clang-tidy.

aminya:
Here it is for llvm:
https://github.com/aminya/setup-cpp/pull/36

lefticus:
> The gcc matcher is only enabled if the compiler is gcc. If the regex is the same as clang, I can add the same thing to clang and clang-tidy.

clang-tidy is automatically getting picked up, since it's not being run as its own compiler here.

it's on my to-do list to have cppcheck output in the appropriate gcc/msvc so it shows up in IDEs as expected, which should also make it get picked up here.

lefticus:
https://github.com/cpp-best-practices/project_options/pull/66 to make sure cppcheck is picked up as expected.

lefticus:
Confirmed @aminya if we merge https://github.com/cpp-best-practices/project_options/pull/66 then we get all of the warnings now picked up and this PR becomes an interesting artifact 

https://github.com/cpp-best-practices/cpp_starter_project/pull/195/files

aminya:
Awesome! 🎉. 

I wish I could add similar support for other CI services like GitLab. I don't know if other services have such similar capabilities.

PR #198, opened 2022-02-17
fix: update project_options to v0.17.0 - add package_project

aminya:


aminya:
@lefticus The way you have added `configured_files` is incorrect. It prevents the project from being packaged. The public include directories should only use relative paths. Using `CMAKE_BINARY_DIR` doesn't work.

lefticus:
Please move this to the https://github.com/cpp-best-practices/cpp_boilerplate_project project so we can stage there, then merge into this project, to keep histories clean and mergeable.

PR #200, opened 2022-03-01
WIP Feature/add non root user in dev container

Jason5480:
Make docker dev container safer to use by logging to a non-root user as described here https://aka.ms/vscode-remote/containers/non-root .
Clean up some defaulted arguments.
Install cmake_format in the container

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -114,6 +114,9 @@ RUN mkdir -p $(include-what-you-use -print-resource-dir 2>/dev/null)
 RUN ln -s $(readlink -f /usr/lib/clang/${LLVM_VER}/include) \
     $(include-what-you-use -print-resource-dir 2>/dev/null)/include
 
+# Install cmake format package
+RUN python3 -m pip install --upgrade pip cmake_format
+
 ## Cleanup cached apt data we don't need anymore
 RUN apt-get autoremove -y && apt-get clean && \
     rm -rf /var/lib/apt/lists/*
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -32,7 +32,7 @@ RUN python3 -m pip install --upgrade pip setuptools && \
 # By default, anything you run in Docker is done as superuser.
 # Conan runs some install commands as superuser, and will prepend `sudo` to
 # these commands, unless `CONAN_SYSREQUIRES_SUDO=0` is in your env variables.
-ENV CONAN_SYSREQUIRES_SUDO 0
+ENV CONAN_SYSREQUIRES_SUDO 1
 # Some packages request that Conan use the system package manager to install
 # a few dependencies. This flag allows Conan to proceed with these installations;
 # leaving this flag undefined can cause some installation failures.
@@ -129,6 +129,8 @@ ENV CXX=${USE_CLANG:+"clang++"}
 # if CC is null, set it to 'gcc' (or leave as is otherwise).
 ENV CC=${CC:-"gcc"}
 ENV CXX=${CXX:-"g++"}
+
+USER $USERNAME
 
 # Include project
 #ADD . /workspaces/cpp_starter_project
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,8 +2,19 @@
 ARG VARIANT="focal"
 FROM ubuntu:${VARIANT}
 
-# Restate the variant to use it later on in the llvm and cmake installations
-ARG VARIANT
+ARG USERNAME="user"
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Create the user
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+    #
+    # [Optional] Add sudo support. Omit if you don't need to install software after connecting.
+    && apt-get update \
+    && apt-get install -y sudo \
+    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
+    && chmod 0440 /etc/sudoers.d/$USERNAME
 
 # Install necessary packages available from standard repos
 RUN apt-get update -qq && export DEBIAN_FRONTEND=noninteractive && \
@@ -40,6 +51,9 @@ RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test && \
 # Set gcc-${GCC_VER} as default gcc
 RUN update-alternatives --install /usr/bin/gcc gcc $(which gcc-${GCC_VER}) 100
 RUN update-alternatives --install /usr/bin/g++ g++ $(which g++-${GCC_VER}) 100
+
+# Restate the variant to use it later on in the llvm and cmake installations
+ARG VARIANT
 
 ARG LLVM_VER="13"
 # Add clang-${LLVM_VER}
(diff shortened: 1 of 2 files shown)


Jason5480:
> Unfortunately, this causes some low level conan package installation to fail since they are dependent on system packages and somehow the installation does not work for non-root users... @ddalcino did you face that also in the past during the very first implementation? `CONAN_SYSREQUIRES_MODE`=enabled doesn't seem to work for regular users.

The errors are not caused by the user changes of this branch. I went in the `main` branch and a new dependency was needed (pkg-config) when I added that, I got the following error

image

I will create a separate issue for that

lefticus:
@Jason5480 I'm currently trying to reorganize things into a light-weight "boilerplate" project and this "starter_project"

Could I trouble you to move this PR over to https://github.com/cpp-best-practices/cpp_boilerplate_project, since that is the new baseline project, and we'll then merge the changes back into the starter_project?

ddalcino:
> Unfortunately, this causes some low level conan package installation to fail since they are dependent on system packages and somehow the installation does not work for non-root users... @ddalcino did you face that also in the past during the very first implementation? `CONAN_SYSREQUIRES_MODE`=enabled doesn't seem to work for regular users.

`CONAN_SYSREQUIRES_MODE`=enabled means that the Conan process does not need superuser privileges, because it already has them. Many Conan packages have dependencies that must be installed via the system package manager, and typically, sudo is required for this. Without `CONAN_SYSREQUIRES_MODE`=enabled, you will need to `apt-get install sudo`. I don't know how to make this work without superuser privileges, unless you stick to packages that don't need it. That means no GUI libraries like SDL.

You could try going through each packages' dependencies manually, and installing them yourself; I've done this in the past, and gotten SDL to build entirely without superuser privileges. It was ridiculous, and I'll never do it again. I do not recommend this approach.

lefticus:
> > Unfortunately, this causes some low level conan package installation to fail since they are dependent on system packages and somehow the installation does not work for non-root users... @ddalcino did you face that also in the past during the very first implementation? `CONAN_SYSREQUIRES_MODE`=enabled doesn't seem to work for regular users.
> 
> `CONAN_SYSREQUIRES_MODE`=enabled means that the Conan process does not need superuser privileges, because it already has them. Many Conan packages have dependencies that must be installed via the system package manager, and typically, sudo is required for this. Without `CONAN_SYSREQUIRES_MODE`=enabled, you will need to `apt-get install sudo`. I don't know how to make this work without superuser privileges, unless you stick to packages that don't need it. That means no GUI libraries like SDL.
> 
> You could try going through each package's dependencies manually, and installing them yourself; I've done this in the past, and gotten SDL to build entirely without superuser privileges. It was ridiculous, and I'll never do it again. I do not recommend this approach.

Thank's @ddalcino for your insight - I really know nothing about container usage in this way

Jason5480:
> @Jason5480 I'm currently trying to reorganize things into a light-weight "boilerplate" project and this "starter_project"
> 
> Could I trouble you to move this PR over to https://github.com/cpp-best-practices/cpp_boilerplate_project, since that is the new baseline project, and we'll then merge the changes back into the starter_project?

I created that one https://github.com/cpp-best-practices/cpp_boilerplate_project/pull/8 where I added some missing format tools into the container. @ddalcino I will reconsider the non-root user solution and I will try to find a solution to work with conan (if any). Using root user in containers is not accepted in many companies for security reasons...

Jason5480:
> @Jason5480 I'm currently trying to reorganize things into a light-weight "boilerplate" project and this "starter_project"
> 
> Could I trouble you to move this PR over to https://github.com/cpp-best-practices/cpp_boilerplate_project, since that is the new baseline project, and we'll then merge the changes back into the starter_project?

I created that one https://github.com/cpp-best-practices/cpp_boilerplate_project/pull/8 where I added some missing format tools to the container. @ddalcino I will reconsider the non-root user solution and I will try to find a solution to work with conan (if any). Using root user in containers in not accepted in many companies for security reasons

Jason5480:
> > Unfortunately, this causes some low level conan package installation to fail since they are dependent on system packages and somehow the installation does not work for non-root users... @ddalcino did you face that also in the past during the very first implementation? `CONAN_SYSREQUIRES_MODE`=enabled doesn't seem to work for regular users.
> 
> `CONAN_SYSREQUIRES_MODE`=enabled means that the Conan process does not need superuser privileges, because it already has them. Many Conan packages have dependencies that must be installed via the system package manager, and typically, sudo is required for this. Without `CONAN_SYSREQUIRES_MODE`=enabled, you will need to `apt-get install sudo`. I don't know how to make this work without superuser privileges, unless you stick to packages that don't need it. That means no GUI libraries like SDL.
> 
> You could try going through each package's dependencies manually, and installing them yourself; I've done this in the past, and gotten SDL to build entirely without superuser privileges. It was ridiculous, and I'll never do it again. I do not recommend this approach.

Do you know any command to check if the current user is a superuser and set those variables accordingly?

ddalcino:
> Do you know any command to check if the current user is a superuser and set those variables accordingly?

You can try checking $EUID == 0, or google it. Personally, I think I would prefer an option that I can set (possibly an ARG) that controls whether or not this runs in superuser mode or not, and set the default to non-superuser. Projects that use SDL etc. would have to either use the option, change the default, or install the dependency outside of Conan.

Thanks for looking into this PR, this is a serious security issue. I hated writing the original in this way, but I didn’t see an alternative.

ddalcino:
> `CONAN_SYSREQUIRES_MODE`=enabled means that the Conan process does not need superuser privileges, because it already has them. Many Conan packages have dependencies that must be installed via the system package manager, and typically, sudo is required for this. Without `CONAN_SYSREQUIRES_MODE`=enabled, you will need to `apt-get install sudo`. I don't know how to make this work without superuser privileges, unless you stick to packages that don't need it. That means no GUI libraries like SDL.

Whoops, looks like I got this backwards. I was thinking of CONAN_SYSREQUIRES_SUDO; see https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_sysrequires_mode#conan-sysrequires-sudo.
CONAN_SYSREQUIRES_MODE is documented here: https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_sysrequires_mode#conan-sysrequires-mode

If I had read the comments I left in the Dockerfile, I would not have made this mistake.

.devcontainer/Dockerfile
```diff
@@ -21,7 +32,7 @@ RUN python3 -m pip install --upgrade pip setuptools && \
 # By default, anything you run in Docker is done as superuser.
 # Conan runs some install commands as superuser, and will prepend `sudo` to
 # these commands, unless `CONAN_SYSREQUIRES_SUDO=0` is in your env variables.
-ENV CONAN_SYSREQUIRES_SUDO 0
+ENV CONAN_SYSREQUIRES_SUDO 1
 # Some packages request that Conan use the system package manager to install
 # a few dependencies. This flag allows Conan to proceed with these installations;
 # leaving this flag undefined can cause some installation failures.
```
ddalcino, line 38:
```suggestion
# See docs at https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_sysrequires_mode#conan-sysrequires-sudo
ENV CONAN_SYSREQUIRES_SUDO 1
# Some packages request that Conan use the system package manager to install
# a few dependencies. This flag allows Conan to proceed with these installations;
# leaving this flag undefined can cause some installation failures.
# See docs at https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_sysrequires_mode#conan-sysrequires-mode
```

Jason5480: I will add this here https://github.com/Jason5480/cpp_boilerplate_project/tree/feature/add_non_root_user_in_docker_container I will close that MR for now as I will not procced any further. I will make a new MR when I find how I can make it work.

PR #212, opened 2022-04-03
Support clang-format

marcpawl:
cmake --build ./build  --target clangformat

Code taken from https://github.com/zemasoft/clangformat-cmake

I am pretty much a cut and past CMake guy, so please look at this closely before accepting.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "clangformat-cmake"]
+	path = clangformat-cmake
+	url = https://github.com/zemasoft/clangformat-cmake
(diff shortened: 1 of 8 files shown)


ddalcino:
We already have this running in CI, triggered on every pull request: https://github.com/cpp-best-practices/cpp_starter_project/blob/main/.github/workflows/auto-clang-format.yml

As far as I can tell, the difference between this PR and what we already have is that this PR gives you an extra build target that runs clang-format. Do we need this?

If a user is concerned that clang-format isn't running often enough, the `auto-clang-format.yml` action can be configured to run on every push, rather than only on pull requests.

marcpawl:
In my case I wanted the target so I can have things formatted prior to making the commit.  For example doing a git diff.  I have users who edit code with vi and formatting gets messed up.  I wanted to have a nice easy way that formats everything for them.

BTW, we are not using GITHUB.

PR #213, opened 2022-04-05
configured files: installing generated header

benthevining:
This adds an example of installing the generated header file, and using generator expressions to use the header in the binary directory when building, or the header in the install tree when linking to the installed project. This also adds an example of using an interface library to express usage requirements.


lefticus:
@benthevining thank you for the PR. Would you mind moving this work over to https://github.com/cpp-best-practices/cpp_boilerplate_project ?

The boilerplate project is intended to be the simpler baseline project, and once changes are made and tested there, they are merged into this project (they share a github history).

Thank you!

